PR #11, opened 2019-11-12
fleet msgs

codebot:
Add fleet-interface messages package. Since it's a new package, it shouldn't cause any issues to merge.


rmf_fleet_msgs/msg/RobotState.msg
```diff
@@ -0,0 +1,6 @@
+string name
+string model
+rmf_fleet_msgs/RobotMode mode
```
mxgrey: Sorry I didn't catch this before it was merged, but I've actually had issues with including the package name when the message is part of the package that's using it. Specifically I think ament doesn't necessarily set the compilation flags up correctly, and you get compiler errors. So I think the package name should just be removed, like:

```suggestion
RobotMode mode
```

codebot: interesting; it seems to compile OK currently (?) but I'll make the changes you suggest, since it results in fewer characters :+1:

mxgrey: Yeah, I don't know exactly what the conditions are, but I think it probably has to do with:

1. The order that the `.msg` files are listed in
2. Race conditions when building with multiple threads

codebot: https://github.com/osrf/rmf_core/pull/12

PR #13, opened 2019-11-14
rmf_door_msgs

codebot:


rmf_door_msgs/msg/DoorRequest.msg
```diff
@@ -0,0 +1,7 @@
+builtin_interfaces/Time request_time
+string requester_id
+string door_name
```
mxgrey: Should we not use `DoorState desired_state` here instead of repeating the same fields?

mxgrey: Oh I guess the `DoorState` has a time field :thinking:

codebot: yeah, we could do that. I just kind of liked how the name of the timestamp made it more obvious that we were defining the time of the request rather than the "internal" time of the door driver.

mxgrey: Yeah, there are fair arguments either way. I'm just afraid of the enums of the door mode drifting because of multiple sources of truth.

rmf_door_msgs/msg/DoorRequest.msg
```diff
@@ -0,0 +1,7 @@
...
+uint8 door_state
```
mxgrey: Would it make sense to call this `door_mode` instead of `door_state` to differentiate it from `DoorState`?

codebot: ah, good idea :+1:

codebot: to only specify the enums in one place, we can push them into a different message

mxgrey: We could do something like

```
# DoorMode.msg
uint8 CLOSED=0
uint8 MOVING=1
uint8 OPEN=2
uint8 value
```

mxgrey: I believe the enums get nested into the class, so in practice it would read as `DoorMode::CLOSED`, `DoorMode::OPEN`, etc.

rmf_door_msgs/msg/DoorRequest.msg
codebot: thanks for the feedback. Incorporated into newer version below.

rmf_door_msgs/msg/DoorRequest.msg
codebot: Thanks for the feedback. Having the enums only in one place is a much better idea. `DoorMode.msg` now included below.

rmf_door_msgs/msg/DoorMode.msg
```diff
@@ -0,0 +1,10 @@
+# The DoorMode message captures the "mode" of an automatic door controller.
+# Most door controllers default to running in "closed" mode, and transition
+# through some sort of "moving" mode until reaching the "open" mode.
+
+uint32 mode
```
mxgrey:
```suggestion
uint32 value
```

This way a user would type `door_state_msg.current_mode.value` instead of `door_state_msg.current_mode.mode`.

codebot: :heavy_check_mark:  nice! I changed the identifier name to `value` and changed the doc strings a bit to reflect that.

PR #29, opened 2019-11-21
Read only fleet adapter

mxgrey:
This introduces a read-only fleet adapter that will send delay and replacement requests to the schedule based on `rmf_fleet_msgs/FleetState` data that comes from the fleet drivers.


mxgrey:
@Yadunund will follow up later with some tests for this adapter.

PR #31, opened 2019-11-21
Fixed tests when compiling with clang

luca-della-vedova:
Required removing auto in favor of Eigen::Vector in all variable declarations of test code.

Now tests all pass even when compiling with clang.


mxgrey:
Thanks for the fix. I tested it on my machine and it's looking happy.

For reference, here's some discussion on the issue: https://www.alecjacobson.com/weblog/?p=4043

Issue #33, opened 2019-11-22
Task ID for msgs in `rmf_task_mgs`


tanyouliang95:
Mainly for enhancement. It will be great if there's a `task_guid` (or any similar name) in both `delivery.msg` and `station.msg`. This will be helpful for the task_initiator (mainly UI) to track each task.

Also, might also want to create a new feedback msg to any task_initiator, something like a  `task_queue.msg` which publishes periodically.

codebot:
Thanks @tanyouliang95  for creating the ticket. I'll make a strawman PR for this where we can discuss further.

tanyouliang95:
Also according to #50 . A list of task states with type `Task.msg` will be published to `/task_states` topic periodically.

codebot:
the topic will be named `/tasks` instead of `/task_states` just to save typing

PR #34, opened 2019-11-22
add task_id and some messages to view task queues

codebot:
Strawman implementation for discussion of #33 to add `task_id` to inbound task requests, and an outbound message called `Tasks` to periodically publish the states of queued, active, recently-completed, and recently-failed tasks.

### DIFF
--- a/rmf_task_msgs/msg/TaskSummary.msg
+++ b/rmf_task_msgs/msg/TaskSummary.msg
@@ -1,0 +1,8 @@
+string task_id
+
+# a brief summary of the current status of the task
+string status
+
+builtin_interfaces/Time submission_time
+builtin_interfaces/Time start_time
+builtin_interfaces/Time end_time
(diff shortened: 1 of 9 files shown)


rmf_task_msgs/msg/TaskSummary.msg
```diff
@@ -0,0 +1,8 @@
+string task_id
+
+# a brief summary of the current status of the task
+string status
+
+builtin_interfaces/Time submission_time
+builtin_interfaces/Time start_time
+builtin_interfaces/Time end_time
```
mxgrey: For tasks that are in progress, should this message be sent with an estimate of the end time?

Or is this message only meant for tasks that have been completed?

mxgrey: Considering the `status` field, I suppose this is probably for in-progress tasks as well, so I guess we should give an end time estimate.

codebot: Yeah the thought/hope was that `end_time` is an estimate for tasks that are in-progress, and it is an "actual" end time for tasks that have completed or failed. I'll add an explanation to the `.msg` file to hopefully make it clear.

PR #36, opened 2019-11-23
Change target_guid to dispenser_name

mxgrey:
I'm not sure what `target_guid` was intended to mean, so I've changed the field name to `dispenser_name`, since that's my best guess at what it was intended to mean.

### DIFF
--- a/rmf_dispenser_msgs/msg/DispenserRequest.msg
+++ b/rmf_dispenser_msgs/msg/DispenserRequest.msg
@@ -1,6 +1,6 @@
 builtin_interfaces/Time time
 string request_guid
-string target_guid
+string dispenser_name
 
 # below are custom workcell message fields
 string transporter_type


tanyouliang95:
FYI. Similarly, in `rmf_dispenser_msgs`, both `DispenserRequest.msg` and `DispenserState.msg` also use `source_guid` and `guid` to represent `dispenser_name`.

mxgrey:
In that case should we consider keeping it as `target_guid` but adding a comment to explain that "target" refers to which dispenser the request is "targeted" at?

tanyouliang95:
Keeping it is okay. However, while looking at other msgs (lift, door, fleet), all use `XXX_name` to represent the "target". It might be good if we keep it consistent. :thinking:

mxgrey:
Okay, let's go with `dispenser_name` for now, and then we can revisit the question to come up with a more consistent scheme after December when we have more time at our disposal.

aaronchongth:
I think that is a great idea, the message field names were kind of still in a prototype stage and should evolve as we find flaws or better alternatives. Thanks for making the necessary changes.

PR #41, opened 2019-11-26
rmf_map_msgs

codebot:
An iteration of our previous collection of map-related messages, incorporating various lessons-learned. Feel free to propose changes!

### DIFF
--- a/rmf_map_msgs/msg/Graph.msg
+++ b/rmf_map_msgs/msg/Graph.msg
@@ -1,0 +1,4 @@
+string name
+GraphNode[] vertices
+GraphEdge[] edges
+Param[] params
--- a/rmf_map_msgs/msg/Lift.msg
+++ b/rmf_map_msgs/msg/Lift.msg
@@ -1,0 +1,2 @@
+string name
+string[] available_floors
--- a/rmf_map_msgs/msg/Lift.msg
+++ b/rmf_map_msgs/msg/Lift.msg
@@ -1,2 +1,4 @@
 string name
-string[] available_floors
+string[] levels
+Door[] doors
+Graph wall_graph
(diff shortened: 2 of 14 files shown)


rmf_map_msgs/msg/Graph.msg
```diff
@@ -0,0 +1,4 @@
+string name
```
koonpeng: I think the map server are no longer required to know about the navigation nodes so all the Graph* messages should be removed?

rmf_map_msgs/msg/Lift.msg
```diff
@@ -0,0 +1,4 @@
+string name
```
koonpeng: adding the lift location at each levels would be nice

rmf_map_msgs/msg/Graph.msg
mxgrey: It looks to me like the intention is for the `Graph.msg` to be used to express

1. The layout of the walls in the environment
2. The layout of each fleet's navigation graph

It's true that we no longer have a single metamap navigation graph, because we're allowing every fleet to have its own navigation graph, but it may still be useful to communicate all the various fleet navigation graphs to the UI so operators have a better idea of what each fleet is doing, can do, or will do.

I don't know exactly what method Morgan has in mind for gathering those navigation graphs, but I could imagine that we may have the fleet adapters report their navigation graphs to the map server when they start up.

rmf_map_msgs/msg/Lift.msg
codebot: Thanks, added in the version forthcoming in a PR in `traffic-editor`

rmf_map_msgs/msg/Graph.msg
codebot: Yes, the intent here is to create a message that contains all of the annotations in `traffic-editor`, so that (for example) a dashboard UI could show the "robot roads," to make it easier for operators to understand the current and near-future behavior of the system.

If this information is not needed and ends up horribly bloating the messages for some applications, we could consider either adding request fields to the `get_building_map` service to exclude things like the navigation-lane or wall graphs, or we could have different service names that would only return the background images or other aspects of interest.

PR #42, opened 2019-11-26
Workcell msgs

aaronchongth:
* added workcell messages

* added some comments on dispenser messages

* considering to use workcell-type state messages in the templates as well


aaronchongth:
* made it more obvious that the workcell-type state, request and result messages have common fields
* added some more comments

mxgrey:
We may want to consider just putting the `rmf_dispenser_msgs` into the `rmf_workcell_msgs` package.

I can't think of any particular reason to keep them separate.

aaronchongth:
cool, that's true, I didn't want to flood message packages like how `rmf_msgs` turned out, but this seems well-contained for it. I'll make the fixes.

PR #43, opened 2019-11-26
merged dispenser messages into workcell messages

aaronchongth:
* merging dispenser into workcell messages

* using the flat convention of message folders


aaronchongth:
that is a very very valid point, and perhaps the biggest flaw to this templated workcell-type concept. It was something I toyed around with, and thought that an adapter library that interacts with `rmf_core` or any other `rmf` core systems using a set of common fields would fall in place with the future `DispatchPlanner` or `ResourceManager` concept.

aaronchongth:
ooohhh, yeah thats the idea! The different types of workcells will be given a whole new set of topics, for example, `Dispenser` will be using `dispenser_states`, `dispenser_requests` and `dispenser_results` by convention. 

Yes having the custom messages hold something like `WorkcellRequest` would work too, but in the same case, the field name will have to be specified if we decide to use a templated workcell adapter library, like 
```
# DispenserRequest.msg

# this field name and type have to be fixed for all types
WorkcellRequest workcell_request 

# all other fields that are specific to this workcell type
float dispenser_online_time
int foo
int bar
```

aaronchongth:
that looks clean and comprehensive, I will work on that in a separate PR as it will require some more migration on `rmf` and `sharp-rmf/rmf-workcell`.

I agree that `WorkcellState` could easily be generic enough for all types. I am however a bit afraid of clogging up a single `workcell_state` topic, unless I use type specific topics as well, like `dispenser_state` with message type `WorkcellState`?

aaronchongth:
icic, sounds good. thanks for the insights! I will close this PR and refer back later.

PR #45, opened 2019-11-29
Refactor delivery message

mxgrey:


tanyouliang95:
Nice that this. Is it better if we can have this 'item' msg in `rmf_workcell_msgs`?

mxgrey:
If it's moved to `rmf_workcell_msgs` would it still be called `DispenserRequestItem`?

Related to that, are there any thoughts on moving everything from `rmf_dispenser_msgs` into `rmf_workcell_msgs`? I don't have strong feelings on this either way.

aaronchongth:
I think we should go ahead and have this inside `rmf_dispenser_msgs`, I have started some migration work on another branch (https://github.com/osrf/rmf_core/tree/common-workcell-msg-scheme), with the intention to migrate `sharp-rmf/rmf-workcell` repo to use `rmf_workcell_msgs` instead of `rmf_msgs`.

The discussions for naming `DispenserRequestItem` would be more relevant then

tanyouliang95:
Yeap, either way will do. Since migration is goin on, we can keep it this way for the time being. :+1:

mxgrey:
I'll merge this soon if there are no objections.

PR #49, opened 2019-12-01
Create a node to supervise the door requests

mxgrey:
Since every fleet adapter is issuing lift and door requests for each of their own robots without consideration for the needs of other fleets, we need to have nodes whose responsibility is to multiplex these requests into a single open or close command.

The strategy used for the door is to gather all the requests that are being emitted by the fleet adapters and keep track of what the latest command was from each adapter for each robot with respect to each door. If the latest request for any combination of `(fleet, robot name)` on a given door is to have the door open, then the door supervisor node will maintain a door open request. When all of the latest requests for a given door were to have it closed, then the door supervisor will issue door close requests.

This makes sure that the open state is favored whenever a robot needs to pass through a door. Otherwise there may be a risk that a fleet adapter may command the door to close on the robot of another fleet's robot that the adapter does not know anything about.

### DIFF
--- a/rmf_fleet_adapter/CMakeLists.txt
+++ b/rmf_fleet_adapter/CMakeLists.txt
@@ -97,6 +97,49 @@ target_include_directories(full_control
 
 # -----------------------------------------------------------------------------
 
+add_executable(lift_supervisor
+  src/lift_supervisor/main.cpp
+  src/lift_supervisor/Node.cpp
+)
+
+target_link_libraries(lift_supervisor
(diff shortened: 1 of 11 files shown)


mxgrey:
This also takes care of monitoring the lift state and issuing the `EmergencyNotice` message when any lift state indicates an emergency has occurred.

For now we are not monitoring lift requests, because that should be done with a higher level of planning intelligence than what is needed for the doors. That feature will be deferred for later.

PR #50, opened 2019-12-02
include the state of the task within TaskSummary

codebot:
instead of having multiple lists within `Tasks.msg`, now that container message is just a single list of `TaskSummary.msg`, each of which has their own state field.

The intent is that when a task is completed it is included with `STATE_COMPLETED` set only for one publication, and then it ceases to be included in this message stream. Similarly for `STATE_FAILED`.


mxgrey:
> when a task is completed it is included with STATE_COMPLETED set only for one publication

Should we be concerned about losing packets and potentially needing to repeat the status of a completed message because a subscriber didn't get the initial message?

PR #68, opened 2020-01-22
Feature/broad phase refinement

Yadunund:
This PR enhances the broad-phase algorithm used for collision checking of trajectories. It iteratively checks for overlap of axis-aligned bounding boxes of trajectory segments with comparable time ranges.

Other additions include 
* `rmf_traffic::geometry::FinalShape::get_characteristic_length()` and modifications to Box.cpp, Circle.cpp, SimplePolygon.cpp, ShapeInternal.cpp
* `rmf_traffic::Spline::get_params()`
* `test_BroadPhase.cpp`
* Formatting changes in `test_Conflict.cpp`

### DIFF
--- a/rmf_traffic/src/rmf_traffic/Conflict.cpp
+++ b/rmf_traffic/src/rmf_traffic/Conflict.cpp
@@ -240,6 +240,108 @@ fcl::ContinuousCollisionRequest make_fcl_request()
   return request;
 }
 
+using BoundingBox = std::pair<Eigen::Vector2d, Eigen::Vector2d>;
+struct Coeffs
+{
+public:
+  double a;
+  double b;
+  double c;
+  double d;
+
+ Coeffs()
+ {
+   a = 0;
+   b = 0;
+   c = 0;
+   d= 0;
+ }
+  Coeffs(
+      double a_,
+      double b_,
+      double c_,
+      double d_)
+  : a(a_),
+    b(b_),
+    c(c_),
+    d(d_)
+  {}
+};
+
+// Container for storing the coefficients of x and y spline motions
+using SplineCoeffs = std::pair<Coeffs, Coeffs>;
+
+SplineCoeffs get_spline_coefficients(
+    Trajectory::Segment& seg_a,
+    Trajectory::Segment& seg_b)
+{
+  SplineCoeffs spline_coeffs;
+
+  auto get_coeffs = [&](int index) -> Coeffs
+  {
+    assert(index > 0 && index < 2);
+    uint64_t initial_time = seg_a.get_finish_time().time_since_epoch().count();
+    uint64_t final_time = seg_b.get_finish_time().time_since_epoch().count();
+    double initial_position = seg_a.get_finish_position()[index];
+    double final_position = seg_b.get_finish_position()[index];
+    double initial_velocity = seg_a.get_finish_velocity()[index];
+    double final_velocity = seg_b.get_finish_velocity()[index];
+
+    auto td = final_time - initial_time;
+    auto x0 = initial_position;
+    auto x1 = final_position;
+    auto v0 = initial_velocity;
+    auto v1 = final_velocity;
+    auto w0 = v0/td;
+    auto w1 = v1/td;
+    
+    auto a = w1 + w0 -2*x1 + 2*x0;
+    auto b = -w1 - 2*w0 +3*x1 -3*x0;
+    auto c = w0;
+    auto d = x0;
+
+    return Coeffs(a, b, c, d);
+  };
+
+  spline_coeffs.first = get_coeffs(0);
+  spline_coeffs.second = get_coeffs(1);
+  return spline_coeffs;
+}
+
+void get_solutions(
+    rmf_traffic::Trajectory::iterator& begin,
+    rmf_traffic::Trajectory::iterator& end,
+    std::vector<double>& x_sols,
+    std::vector<double>& y_sols,
+    std::mutex* mutex = nullptr)
+{
+  for (auto it = begin; it != end; it++)
+  {
+
+  }
+}
+
+BoundingBox get_bounding_box(const Trajectory& trajectory)
+{
+  BoundingBox bounding_box;
+  std::vector<double> x_sols;
+  std::vector<double> y_sols;
+  auto size = trajectory.size();
+  std::mutex mutex;
+
+  // get x_sols and y_sols
+
+  Eigen::Vector2d min_coord = Eigen::Vector2d{
+      *std::min_element(x_sols.begin(), x_sols.end()),
+      *std::min_element(y_sols.begin(), y_sols.end())};
+  Eigen::Vector2d max_coord = Eigen::Vector2d{
+      *std::max_element(x_sols.begin(), x_sols.end()),
+      *std::max_element(y_sols.begin(), y_sols.end())};
+  bounding_box.first = min_coord;
+  bounding_box.second = max_coord;
+  return bounding_box;
+}
+
 } // anonymous namespace
 
 class DetectConflict::Implementation
--- a/rmf_traffic/src/rmf_traffic/Conflict.cpp
+++ b/rmf_traffic/src/rmf_traffic/Conflict.cpp
@@ -21,7 +21,6 @@
 #include "StaticMotion.hpp"
 
 #include <rmf_traffic/Conflict.hpp>
-#include <rmf_traffic/geometry/Circle.hpp>
 
 #include <fcl/continuous_collision.h>
 #include <fcl/ccd/motion.h>
--- a/rmf_traffic/src/rmf_traffic/Conflict.cpp
+++ b/rmf_traffic/src/rmf_traffic/Conflict.cpp
@@ -126,114 +126,37 @@ std::vector<ConflictData> DetectConflict::between(
 namespace {
 
 using BoundingBox = std::pair<Eigen::Vector2d, Eigen::Vector2d>;
-struct Coeffs
-{
-public:
-  double a;
-  double b;
-  double c;
-  double d;
-  double ti;
-  double tf;
-
-  Coeffs()
-  {
-    a = 0;
-    b = 0;
-    c = 0;
-    d = 0;
-    ti = 0;
-    tf = 0;
-  }
-
-  Coeffs(
-      double a_,
-      double b_,
-      double c_,
-      double d_,
-      double ti_,
-      double tf_)
-  : a(a_),
-    b(b_),
-    c(c_),
-    d(d_),
-    ti(ti_),
-    tf(tf_)
-  {}
-};
-
-// Container for storing the coefficients of x and y spline motions
-using SplineCoeffs = std::pair<Coeffs, Coeffs>;
-
-// TODO Add get_param() function to Spline.hpp and delete below function
-SplineCoeffs get_spline_coefficients(
-    rmf_traffic::Trajectory::const_iterator seg_a,
-    rmf_traffic::Trajectory::const_iterator seg_b)
-{
-  SplineCoeffs spline_coeffs;
-
-  auto get_coeffs = [&](int index) -> Coeffs
-  {
-    assert(index >= 0 && index < 2);
-    // Time in seconds and positions in meters
-    double ti = seg_a->get_finish_time().time_since_epoch().count() / 1E9;
-    double tf = seg_b->get_finish_time().time_since_epoch().count() / 1E9;
-    double initial_position = seg_a->get_finish_position()[index];
-    double final_position = seg_b->get_finish_position()[index];
-    double initial_velocity = seg_a->get_finish_velocity()[index];
-    double final_velocity = seg_b->get_finish_velocity()[index];
-
-    double td = (tf - ti);
-    double x0 = initial_position;
-    double x1 = final_position;
-    double v0 = initial_velocity;
-    double v1 = final_velocity;
-    double w0 = v0/td;
-    double w1 = v1/td;
-    
-    double a = w1 + w0 -2*x1 + 2*x0;
-    double b = -w1 - 2*w0 +3*x1 -3*x0;
-    double c = w0;
-    double d = x0;
-
-    return Coeffs(a, b, c, d, ti, tf);
-  };
-
-  spline_coeffs.first = get_coeffs(0);
-  spline_coeffs.second = get_coeffs(1);
-
-  return spline_coeffs;
-}
 
 double evaluate_spline(
-    const Coeffs coeffs,
+    const Eigen::Vector4d& coeffs,
     const double t)
 {
   // Assume time is parameterized [0,1]
-  return (coeffs.a * t * t * t
-      + coeffs.b * t * t
-      + coeffs.c * t
-      + coeffs.d);
+  return (coeffs[3] * t * t * t
+      + coeffs[2] * t * t
+      + coeffs[1] * t
+      + coeffs[0]);
 }
 
 void get_local_extrema(
-    const Coeffs& coeffs,
-    std::vector<double>& sols)
+    const Eigen::Vector4d& coeffs,
+    std::array<double, 2>& extrema)
 {
+  std::vector<double> t_sols;
   // Store boundary values as potential extrema
-  sols.emplace_back(evaluate_spline(coeffs, 0));
-  sols.emplace_back(evaluate_spline(coeffs, 1));
+  t_sols.emplace_back(evaluate_spline(coeffs, 0));
+  t_sols.emplace_back(evaluate_spline(coeffs, 1));
 
   // When derivate of spline motion is not quadratic
-  if (coeffs.a == 0)
+  if (coeffs[3] == 0)
   {
-    if (coeffs.b != 0)
-      sols.emplace_back(-coeffs.c / coeffs.b);
+    if (coeffs[2] != 0)
+      t_sols.emplace_back(-coeffs[1] / coeffs[2]);
   }
   else
   {
     // Calculate the discriminant otherwise
-    double D = (4 * coeffs.b * coeffs.b) - 12 * coeffs.a * coeffs.c;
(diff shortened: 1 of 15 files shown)


rmf_traffic/src/rmf_traffic/Conflict.cpp
```diff
@@ -121,54 +121,116 @@ std::vector<ConflictData> DetectConflict::between(
 }
 
 //==============================================================================
-bool DetectConflict::broad_phase(
-    const Trajectory& trajectory_a,
-    const Trajectory& trajectory_b)
+
+namespace {
+
+using BoundingBox = std::pair<Eigen::Vector2d, Eigen::Vector2d>;
```
mxgrey: I recommend defining structs with explicit fields, because it's hard to remember what each index means or what first/second refers to. For example:

```
struct BoundingBox
{
  Eigen::Vector2d min;
  Eigen::Vector2d max;
};
```

rmf_traffic/src/rmf_traffic/Conflict.cpp
```diff
@@ -121,54 +121,116 @@ std::vector<ConflictData> DetectConflict::between(
...
+  };
+
+  Eigen::Vector2d centre_a = centre(box_a);
+  Eigen::Vector2d centre_b = centre(box_b);
+  double width_a = std::abs(box_a.second[0] - box_a.first[0]);
+  double height_a = std::abs(box_a.second[1] - box_a.first[1]);
+  double width_b = std::abs(box_b.second[0] - box_b.first[0]);
+  double height_b = std::abs(box_b.second[1] - box_b.first[1]);
+
+  return std::abs(centre_b[0] - centre_a[0]) < 0.5 * (width_a + width_b)
```
mxgrey: Maybe I'm missing something, but why not just compare the extrema? I.e.:

```
for (std::size_t i=0; i < 2; ++i)
{
  if (box_a.second[i] < box_b.first[i])
    return false;

  if (box_b.second[i] < box_a.first[i])
    return false;
}

return true;
```

rmf_traffic/src/rmf_traffic/Conflict.cpp
```diff
@@ -121,54 +121,116 @@ std::vector<ConflictData> DetectConflict::between(
...
-    // cannot be any conflict.
-    return false;
+      t_sols.emplace_back(extrema1);
+      t_sols.emplace_back(extrema2);
+    }
   }
 
-  const auto* t_b0 = trajectory_b.start_time();
-  const auto* t_af = trajectory_a.finish_time();
+  extrema[0] = *std::min_element(t_sols.begin(), t_sols.end());
```
mxgrey: I recommend adding `assert(!t_sols.empty());` before this line just as a precaution.

rmf_traffic/test/unit/test_BroadPhase.cpp
```diff
@@ -0,0 +1,302 @@
...
+    if (test_performance)
+    {
+      const double sec = rmf_traffic::time::to_seconds(end_time - start_time);
+      std::cout << "Non-overlapping straight-line trajectories with " << n_seg << " segments\n";
+      std::cout << "Total: " << sec << std::endl;
+      std::cout << "Per run: " << sec/N << std::endl;
+    }
+  }
+
+  GIVEN("Time ranges and bounding boxes overlap but segments do not intersect")
```
mxgrey: What about this same test but the time of `trajectory_b` is offset just enough that the bounding boxes don't intersect when time is accounted for?

rmf_traffic/src/rmf_traffic/Conflict.cpp
```diff
@@ -121,54 +121,116 @@ std::vector<ConflictData> DetectConflict::between(
...
+using BoundingBox = std::pair<Eigen::Vector2d, Eigen::Vector2d>;
+
+double evaluate_spline(
+    const Eigen::Vector4d& coeffs,
+    const double t)
 {
-  std::size_t min_size = std::min(trajectory_a.size(), trajectory_b.size());
-  if(min_size < 2)
+  // Assume time is parameterized [0,1]
+  return (coeffs[3] * t * t * t
```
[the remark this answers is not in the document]
Yadunund: That was exactly the misconception I had. But when I switch to the loop, interestingly I do get slightly slower performance as seen in the screenshots below (maybe not significant). I'm building with `--cmake-args -DCMAKE_BUILD_TYPE=RELEASE`.
Current:
image

With loop:
image

rmf_traffic/src/rmf_traffic/Conflict.cpp
```diff
@@ -121,54 +121,116 @@ std::vector<ConflictData> DetectConflict::between(
...
+    const Eigen::Vector4d& coeffs,
+    std::array<double, 2>& extrema)
+{
+  std::vector<double> t_sols;
+  // Store boundary values as potential extrema
+  t_sols.emplace_back(evaluate_spline(coeffs, 0));
+  t_sols.emplace_back(evaluate_spline(coeffs, 1));
+
+  // When derivate of spline motion is not quadratic
+  if (coeffs[3] == 0)
```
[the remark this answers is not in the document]
Yadunund: Right, thank you for highlighting this. Resolved.

rmf_traffic/src/rmf_traffic/Conflict.cpp
```diff
@@ -121,54 +121,116 @@ std::vector<ConflictData> DetectConflict::between(
...
   {
-    throw invalid_trajectory_error::Implementation
-        ::make_segment_num_error(min_size);
+    if (coeffs[2] != 0)
```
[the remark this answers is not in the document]
Yadunund: Resolved.

rmf_traffic/src/rmf_traffic/Conflict.cpp
```diff
@@ -121,54 +121,116 @@ std::vector<ConflictData> DetectConflict::between(
...
-  if(trajectory_a.get_map_name() != trajectory_b.get_map_name())
-    return false;
-
-  const auto* t_a0 = trajectory_a.start_time();
-  const auto* t_bf = trajectory_b.finish_time();
+    if (D < 0)
+    {
+      // Do nothing
+    }
+    else if (D == 0)
```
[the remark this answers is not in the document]
Yadunund: Got it, thank you for pointing this out.

rmf_traffic/src/rmf_traffic/Conflict.cpp
```diff
@@ -121,54 +121,116 @@ std::vector<ConflictData> DetectConflict::between(
...
   {
-    throw invalid_trajectory_error::Implementation
-        ::make_segment_num_error(min_size);
+    if (coeffs[2] != 0)
+      t_sols.emplace_back(-coeffs[1] / coeffs[2]);
   }
+  else
+  {
+    // Calculate the discriminant otherwise
+    double D = (4 * coeffs[2] * coeffs[2]) - 12 * coeffs[3] * coeffs[1];
```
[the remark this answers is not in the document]
Yadunund: Noted!

rmf_traffic/src/rmf_traffic/Conflict.cpp
```diff
@@ -121,54 +121,116 @@ std::vector<ConflictData> DetectConflict::between(
...
 
-  if(trajectory_a.get_map_name() != trajectory_b.get_map_name())
-    return false;
-
-  const auto* t_a0 = trajectory_a.start_time();
-  const auto* t_bf = trajectory_b.finish_time();
+    if (D < 0)
+    {
+      // Do nothing
```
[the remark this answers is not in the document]
Yadunund: Noted!

rmf_traffic/src/rmf_traffic/Conflict.cpp
```diff
@@ -121,54 +121,116 @@ std::vector<ConflictData> DetectConflict::between(
...
+  // Assume time is parameterized [0,1]
+  return (coeffs[3] * t * t * t
+      + coeffs[2] * t * t
+      + coeffs[1] * t
+      + coeffs[0]);
+}
+
+void get_local_extrema(
+    const Eigen::Vector4d& coeffs,
+    std::array<double, 2>& extrema)
```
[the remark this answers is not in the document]
Yadunund: I was not aware of this optimization by the complier. Thank you for letting me know. I have changed `extrema` to return by value.

rmf_traffic/src/rmf_traffic/Conflict.cpp
Yadunund: This greatly improves readability of the code! Thanks for the suggestion.

rmf_traffic/src/rmf_traffic/Conflict.cpp
Yadunund: I did originally compare the extrema but in an inefficient (slower performance time) way as below
```
     return (box_a.second[0] > box_b.first[0] &&
    box_a.first[0] < box_b.second[0] &&
    box_a.second[1] > box_b.first[1] &&
    box_a.first[1] < box_b.second[1]);
```
You suggestion to return false as soon as a check fails is great and more performant. I have implemented this.

rmf_traffic/src/rmf_traffic/Conflict.cpp
Yadunund: Noted.

rmf_traffic/src/rmf_traffic/geometry/SimplePolygon.cpp
```diff
@@ -655,9 +655,10 @@ void SimplePolygon::insert_point(
 //==============================================================================
 FinalShape SimplePolygon::finalize() const
 {
+  //TODO calculate the characteristic length for SimplePolygon
```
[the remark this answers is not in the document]
Yadunund: Noted. Implemented!

rmf_traffic/src/rmf_traffic/Conflict.cpp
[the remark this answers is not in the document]
mxgrey: Ah wow. After some more reading, it seems that it may be the case that `pow(~)` only gets optimized when the power is an integer literal. That may explain the loss of performance. In theory that should mean that unwrapping the loop and still using `pow(~)` might perform equivalently, but I guess at that point we're not gaining much.

Thanks for benchmarking this!

rmf_traffic/test/unit/test_BroadPhase.cpp
Yadunund: The test label is not clear but the `time ranges` and `bounding boxes` referred to are of the entire trajectories. The segments are already defined such that their bounding boxes do not intersect when time is accounted for. I could make the time differences more apparent perhaps?

mxgrey: Oh okay! Yep, after looking at the test more carefully, it does seem to be doing what I was suggesting, thanks!

Maybe we could rephrase the test description to something like "Bounding boxes overlap in space, but are offset in time"?

rmf_traffic/src/rmf_traffic/Conflict.cpp
```diff
@@ -121,54 +121,116 @@ std::vector<ConflictData> DetectConflict::between(
...
+  t_sols.emplace_back(evaluate_spline(coeffs, 0));
+  t_sols.emplace_back(evaluate_spline(coeffs, 1));
+
+  // When derivate of spline motion is not quadratic
+  if (coeffs[3] == 0)
   {
-    throw invalid_trajectory_error::Implementation
-        ::make_segment_num_error(min_size);
+    if (coeffs[2] != 0)
+      t_sols.emplace_back(-coeffs[1] / coeffs[2]);
```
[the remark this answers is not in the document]
Yadunund: Whoops. Missed this one in the previous push. Thanks for catching the error.

rmf_traffic/test/unit/test_BroadPhase.cpp
Yadunund: That's a much better label, will update accordingly. Thanks!

rmf_traffic/src/rmf_traffic/Conflict.cpp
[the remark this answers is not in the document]
Yadunund: Actually, I think it should be
`t_sols.emplace_back(evaluate_spline(coeffs, -coeffs[1]/(2*coeffs[2]));`

rmf_traffic/src/rmf_traffic/geometry/Box.cpp
```diff
@@ -98,17 +98,23 @@ double Box::get_y_length() const
 //==============================================================================
 FinalShape Box::finalize() const
 {
+  double characteristic_length = 0.5 * std::sqrt(
```
Yadunund: I modified the characteristic length for Box to be equal to half the diagonal. Would that be more appropriate?

rmf_traffic/src/rmf_traffic/Conflict.cpp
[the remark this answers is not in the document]
mxgrey: Ah yep, that's the actual correct value!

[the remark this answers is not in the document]
mxgrey: Unless wait.... not if it's an array of the time values :thinking:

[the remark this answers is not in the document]
mxgrey: It looks like `t_sols` should be renamed to something more like `extrema_candidates`.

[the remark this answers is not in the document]
Yadunund: Sorry for the poor nomenclature. Yeah that's much better.

rmf_traffic/src/rmf_traffic/geometry/Box.cpp
mxgrey: Ah yeah, I think that's a more accurate characteristic length.

rmf_traffic/src/rmf_traffic/Conflict.cpp
```diff
@@ -135,38 +139,42 @@ double evaluate_spline(
       + coeffs[2] * t * t
       + coeffs[1] * t
       + coeffs[0]);
+  // double value = 0.0;
```
mxgrey: If we're going to keep this in as a comment, let's add a remark about how the performance of this loop approach was profiled to be worse than the long version above.

rmf_traffic/src/rmf_traffic/geometry/SimplePolygon.cpp
```diff
@@ -655,10 +655,16 @@ void SimplePolygon::insert_point(
 //==============================================================================
 FinalShape SimplePolygon::finalize() const
 {
-  //TODO calculate the characteristic length for SimplePolygon
+  double characteristic_length = 0;
+  for (const Eigen::Vector2d& point : this->get_points())
+  {
+    double distance = (point - Eigen::Vector2d{0, 0}).norm();
```
mxgrey: This is equivalent to `const double distance = point.norm();`

rmf_traffic/test/unit/test_BroadPhase.cpp
```diff
@@ -18,6 +18,8 @@
 #include <rmf_utils/catch.hpp>
 
 #include <rmf_traffic/Conflict.hpp>
+#include <rmf_traffic/geometry/SimplePolygon.hpp>
```
mxgrey: It looks like the `SimplePolygon` header was added to the test file but never actually tested.

rmf_traffic/src/rmf_traffic/Conflict.cpp
```diff
@@ -121,55 +121,121 @@ std::vector<ConflictData> DetectConflict::between(
...
+  std::array<double, 2> extrema_x = get_local_extrema(params.coeffs[0]);
+  std::array<double, 2> extrema_y =  get_local_extrema(params.coeffs[1]);
+
+  Eigen::Vector2d min_coord = Eigen::Vector2d{extrema_x[0], extrema_y[0]};
+  Eigen::Vector2d max_coord = Eigen::Vector2d{extrema_x[1], extrema_y[1]};
+
+  double char_length =  params.profile_ptr->get_shape()
+      ->get_characteristic_length();
+
+  if (char_length > 0)
```
mxgrey: How about replacing this if-statement with `assert(char_length >= 0.0);`?

If we ever come across a negative characteristic length, that's a bug. But a characteristic length of zero could be valid for a "particle" shape.

Yadunund: Got it. Have update accordingly.

rmf_traffic/src/rmf_traffic/geometry/SimplePolygon.cpp
Yadunund: Noted.

rmf_traffic/src/rmf_traffic/Conflict.cpp
Yadunund: I think the comments have been removed in the latest version of the file.

rmf_traffic/test/unit/test_BroadPhase.cpp
Yadunund: Thanks for catching that. I've put the tests for characteristic lengths in `make_shapes.cpp`

Jason-Zhou1:
Hi @Yadunund, 
Would you please briefly describe how to detect the trajectory conflict? Which files have been involved? I have found DetectConflict.cpp but not Conflict.cpp file in the rmf_traffic directory.

Jason-Zhou1:
Okay, thank you @codebot.

mxgrey:
> I have found DetectConflict.cpp but not Conflict.cpp file in the rmf_traffic directory. Are they the same?

That file was mostly just renamed from `Conflict.cpp` to `DetectConflict.cpp`. Nothing about the implementation has really changed, except we have a smarter broadphase now.

PR #77, opened 2020-02-17
Schedule Participants

mxgrey:
#### Note: This branch does not currently compile, but it needs to be reviewed before it gets any bigger, so we are targeting it at the `develop` branch.

This PR is the first step in introducing a new participant-based scheduling system. This is a complete redesign of the previous traffic scheduling system, but with the same overall objective.

The new design is motivated by experimental results which revealed the inherent fragility of a distributed scheduling system that also tries to be as light on network traffic as possible. The initial proof-of-concept design was transaction-based, where every modification to the schedule produced a receipt for the modifier, and that receipt could be used to make additional modifications later. The goal was to allow small changes to the schedule (e.g. delays) to be communicated very efficiently. However, the complexity in the dynamics of schedule IDs meant that a small error in transmission could completely derail the coherence of the schedule across all the modifiers and mirrors.

To deal with that, we are introducing the concept of a "schedule participant" which formalizes the idea of a remote agent who is able to make modifications to the schedule. Now modifications to the schedule are compartmentalized between different participants. Modifications to the schedule are no longer transactional: no receipt is given back when a modification is made. Instead, each participant is responsible for tracking and declaring the sequence of modifications that it is making. The basic operational outline is like this:

* A participant registers itself with the schedule database and receives a `ParticipantId` (this is a one-time transaction per participant which will be performed using a ROS2 service)
* When a participant makes a change, the message will include its `ParticipantId`
* The message will also include an `ItineraryVersion` which is a sequence number (independent between participants) for modifications to the schedule. The schedule database expects `ItineraryVersion` values to increase by 1 with each subsequent message per participant. If the schedule receives a message whose `ItineraryVersion` does not immediately follow the last one that was received, then the database will report an `Inconsistency` for that participant.
* If a participant sees that the schedule database is reporting an inconsistency for it, then it can retransmit the change.

With this scheme, every potential error is compartmentalized for each participant, so that unrelated schedule participants do not pose a threat to other participants. There is also a clear and easy recovery option in the form of "nullifying changes".

A **nullifying change** is a category of schedule change that nullifies the history of changes for a participant. There are two types of nullifying change: 

* `set` which replaces the old itinerary of the participant with a completely new itinerary, and
* `erase itinerary` which simply erases whatever the current itinerary is for the participant

In contrast there are three types of **incremental changes**, which represent changes that build on top of the previous changes of a participant. These include:

* `extend` which adds new routes to the existing itinerary of a participant
* `delay` which introduces a delay to the current itinerary of a participant
* `erase route` which erases one of the routes inside a participant's current itinerary

If the incremental changes are received out of order or one of them is missing, the schedule might severely misrepresent the itinerary of the participant. Therefore, the changes won't be applied out of order. If any items in an incremental change sequence are missing, an `Inconsistency` report will be triggered for the participant. The participant can then choose to retransmit the incremental changes that are missing, or it may transmit a new nullifying change to fix the inconsistency. Either will be accepted by the schedule database.

### Current Status

So far everything is (re)designed and (re)implemented for `Viewer`, `Database`, `Mirror`, `Change`, and `Patch`. There are also very considerable changes to `Trajectory` to make the `Trajectory` class more pure and decouple it from map and robot mode information.

The next step will be to make changes to the parts of `rmf_traffic` that have been negatively impacted by the API changes, especially the tests. After that, changes will be needed to downstream packages like `rmf_traffic_ros2` and `rmf_fleet_adapter`. These will occur in future pull requests.

Considering how extensive the changes are in this PR, I recommend using GitHub's `Split` view and just looking at the green (addition) text, and generally ignore the red text.


rmf_traffic/include/rmf_traffic/Trajectory.hpp
```diff
@@ -43,270 +42,115 @@ class Trajectory
...
 
-    /// Change the finish time of this Trajectory Segment. Note that this
-    /// function will only affect this segment, and may cause this Segment to be
+    /// Change the finish time of this Trajectory Waypoint. Note that this
+    /// function will only affect this waypoint, and may cause this waypoint to be
     /// reordered within the Trajectory.
     ///
-    /// To change the finish time for this segment while preserving the relative
+    /// To change the finish time for this waypoint while preserving the relative
     /// times of all subsequent Trajectory segments, use adjust_finish_times()
```
Yadunund: Should this also be waypoint? 
```suggestion
    /// times of all subsequent Trajectory **waypoints**, use adjust_finish_times()
```

rmf_traffic/include/rmf_traffic/Trajectory.hpp
```diff
@@ -43,270 +42,115 @@ class Trajectory
...
     /// \param[in] new_time
-    ///   The new finishing time for this Trajectory Segment.
+    ///   The new finishing time for this Trajectory Waypoint.
     ///
     /// \sa adjust_finish_times(Time new_time)
-    Segment& set_finish_time(Time new_time);
+    Waypoint& change_time(Time new_time);
 
-    /// Adjust the finishing time of this segment and all subsequent segments by
+    /// Adjust the finishing time of this waypoint and all subsequent segments by
```
Yadunund: Same as above.
```suggestion
    /// Adjust the finishing time of this waypoint and all subsequent waypoints by
```

rmf_traffic/include/rmf_traffic/Trajectory.hpp
```diff
@@ -43,270 +42,115 @@ class Trajectory
...
-  // Trajectory::Segment information.
+  // Trajectory::Waypoint information.
   template<typename SegT>
   class base_iterator;
-  using iterator = base_iterator<Segment>;
-  using const_iterator = base_iterator<const Segment>;
+  using iterator = base_iterator<Waypoint>;
+  using const_iterator = base_iterator<const Waypoint>;
 
   /// Create a Trajectory that takes place on the specified map
```
Yadunund: Given that a trajectory is now decoupled from any map
```suggestion
  /// Create an empty Trajectory
```

rmf_traffic/include/rmf_traffic/Trajectory.hpp
mxgrey: I'm thinking we should also eliminate the use of "finishing" when referring to the time value of a waypoint. Now the trajectory elements simply represent waypoints instead of continuous segments, so there's no longer a point in qualifying the values with "finishing".

rmf_traffic/include/rmf_traffic/Trajectory.hpp
```diff
@@ -43,270 +42,115 @@ class Trajectory
...
-    /// Segment's finish time to be less than or equal to the finish time of its
-    /// preceding Segment, or else a std::invalid_argument exception will be
+    /// Waypoint's finish time to be less than or equal to the finish time of its
+    /// preceding Waypoint, or else a std::invalid_argument exception will be
     /// thrown.
     ///
     /// \param[in] delta_t
-    ///   How much to change the finishing time of this segment and all later
+    ///   How much to change the finishing time of this waypoint and all later
     ///   segments. If negative, it must not cross over the finish time of the
```
Yadunund:
```suggestion
    ///   waypoints. If negative, it must not cross over the finish time of the
```

rmf_traffic/include/rmf_traffic/Trajectory.hpp
```diff
@@ -43,270 +42,115 @@ class Trajectory
...
-    void adjust_finish_times(Duration delta_t);
-
-    /// Compute the motion across this Trajectory segment
-    std::unique_ptr<Motion> compute_motion() const;
+    void adjust_times(Duration delta_t);
 
     class Implementation;
   private:
 
     /// \internal Private constructor. Use Trajectory::add_segment() to create
```
Yadunund: Does the API contain this function? If so perhaps it should also be renamed
```suggestion
    /// \internal Private constructor. Use Trajectory::add_waypoint() to create
```

mxgrey: Ah that's a very very outdated comment. The function should be `Trajectory::insert(~)` now.

rmf_traffic/test/unit/test_Trajectory.cpp
```diff
@@ -165,35 +165,35 @@ SCENARIO("Segment Unit Tests")
...
     {
       rmf_traffic::Trajectory trajectory{"test_map"};
       auto result = trajectory.insert(time, guided_unitbox_profile, pos, vel);
-      const rmf_traffic::Trajectory::Segment &segment = *(result.it);
+      const rmf_traffic::Trajectory::Waypoint &waypoint = *(result.it);
 
       *guided_unitbox_profile = *queued_unitCircle_profile;
 
-      THEN("Segment profile is updated.")
+      THEN("Waypoint profile is updated.")
```
Yadunund: This test is no longer relevant as waypoints no longer store profiles.

rmf_traffic/test/unit/test_Trajectory.cpp
```diff
@@ -242,67 +242,67 @@ SCENARIO("Segment Unit Tests")
     inputs.push_back({time + 20s, UnitBox, Eigen::Vector3d(2, 2, 2), Eigen::Vector3d(0, 0, 0)});
     rmf_traffic::Trajectory trajectory = create_test_trajectory(inputs);
     rmf_traffic::Trajectory::iterator trajectory_it = trajectory.begin();
-    rmf_traffic::Trajectory::Segment &segment = *trajectory_it;
-    rmf_traffic::Trajectory::Segment &segment_10s = *(++trajectory_it);
+    rmf_traffic::Trajectory::Waypoint &waypoint = *trajectory_it;
+    rmf_traffic::Trajectory::Waypoint &segment_10s = *(++trajectory_it);
 
     WHEN("Setting a new profile using set_profile function")
```
Yadunund: This test is also no longer relevant.
```suggestion
    WHEN("Setting a new profile using set_profile function")
```

rmf_traffic/test/unit/test_Trajectory.cpp
```diff
@@ -752,29 +752,29 @@ SCENARIO("Trajectory and base_iterator unit tests")
...
 
-    WHEN("Finding a segment at the precise time specified")
+    WHEN("Finding a waypoint at the precise time specified")
     {
-      THEN("Segment is retreieved successfully")
+      THEN("Waypoint is retreieved successfully")
       {
-        CHECK(trajectory.find(time)->get_finish_position() == Eigen::Vector3d(0, 0, 0));
+        CHECK(trajectory.find(time)->position() == Eigen::Vector3d(0, 0, 0));
         CHECK(trajectory.find(time + 10s)->get_finish_position() == Eigen::Vector3d(2, 2, 2));
```
Yadunund:
```suggestion
        CHECK(trajectory.find(time + 10s)->position() == Eigen::Vector3d(2, 2, 2));
```

rmf_traffic/src/rmf_traffic/Spline.cpp
```diff
@@ -67,18 +67,18 @@ Spline::Parameters compute_parameters(
...
-  const Eigen::Vector3d x0 = start.get_finish_position();
-  const Eigen::Vector3d x1 = finish.get_finish_position();
-  const Eigen::Vector3d v0 = delta_t * start.get_finish_velocity();
-  const Eigen::Vector3d v1 = delta_t * finish.get_finish_velocity();
+  const Eigen::Vector3d x0 = start.position();
+  const Eigen::Vector3d x1 = finish.position();
+  const Eigen::Vector3d v0 = delta_t * start.velocity();
+  const Eigen::Vector3d v1 = delta_t * finish.velocity();
 
   const rmf_traffic::Trajectory::ConstProfilePtr profile_ptr =
```
Yadunund: This was added in PR #68 to access the `characteristic_length` of the profile's shape when computing the bounding box for braodpahse collision detection. As the trajectory iterators no longer contain profiles, this would result in an error. I guess `Spline::Parameters` should no longer contain the `profile_ptr` element. We will need to modify `DetectConflict::broad_phase` suitably.

mxgrey: Yep, I'm doing a redesign/refactor of `Collision.hpp` right now.

rmf_traffic/src/rmf_traffic/Trajectory.cpp
```diff
@@ -513,9 +346,9 @@ void Trajectory::Segment::adjust_finish_times(Duration delta_t)
       const auto tc = (new_time).time_since_epoch().count();
 
       const std::string error =
-          std::string("[Trajectory::Segment::adjust_finish_times] ")
+          std::string("[Trajectory::Waypoint::adjust_finish_times] ")
```
Yadunund:
```suggestion
          std::string("[Trajectory::Waypoint::adjust_times] ")
```

rmf_traffic/src/rmf_traffic/Trajectory.cpp
```diff
@@ -539,24 +372,24 @@ void Trajectory::Segment::adjust_finish_times(Duration delta_t)
...
+  for(internal::WaypointList::iterator& it : list_iterators)
   {
     const Time new_time = it->data.finish_time;
     ordering.emplace_hint(ordering.end(), new_time, std::move(it));
   }
 }
 
 //==============================================================================
-std::unique_ptr<Motion> Trajectory::Segment::compute_motion() const
+std::unique_ptr<Motion> Trajectory::Waypoint::compute_motion() const
```
Yadunund: The declaration of this function was removed from `Trajectory.hpp`.

rmf_traffic/include/rmf_traffic/schedule/Viewer.hpp
```diff
@@ -80,25 +81,33 @@ class Viewer
...
+  // and participant descriptions.
 
-  /// Get the latest version number of this Database.
-  Version latest_version() const;
+  /// Get the set of active participant IDs.
+  virtual const std::unordered_set<ParticipantId>& participant_ids() const = 0;
+
+  /// Get the information of the specified participant if it is available.
+  /// If a participant with the specified ID is not registered with the
+  /// schedule, then this will return a nullopt.
```
Yadunund:
```suggestion
  /// schedule, then this will return a nullptr.
```
or should the return type be rmf_utils::optional<ParticipantDescription> instead?

rmf_traffic/include/rmf_traffic/schedule/Writer.hpp
```diff
@@ -0,0 +1,168 @@
...
+  virtual void delay(
+      ParticipantId participant,
+      Time from,
+      Duration delay,
+      ItineraryVersion version) = 0;
+
+  /// Erase an itinerary from this database.
+  ///
+  /// \param[in] participant
+  ///   The ID of the participant whose itinerary is being delayed.
```
aaronchongth: minor documentation typo, `delayed` should be `erased` for this method

rmf_traffic/include/rmf_traffic/schedule/Writer.hpp
```diff
@@ -0,0 +1,168 @@
...
+  ///   The version for this itinerary change
+  ///
+  virtual void erase(
+      ParticipantId participant,
+      ItineraryVersion version) = 0;
+
+  /// Erase a route from an itinerary.
+  ///
+  /// \param[in] participant
+  ///   The ID of the participant whose itinerary is being delayed.
```
aaronchongth: same here, `delayed` should be `erased`

rmf_traffic/src/rmf_traffic/schedule/InconsistencyTracker.hpp
```diff
@@ -0,0 +1,138 @@
...
+    std::function<void()>& _callback;
+    bool _set = false;
+  };
+
+  /// Check whether an entry has an inconsistency
+  ///
+  /// \param[in] version
+  ///   The declared version of this change
+  ///
+  /// \param[in] inconsistencies
```
Yadunund: The function signature does not contain an attribute for "current record of inconsistencies". I might be misunderstanding here but is this information is contained in `_ranges`?

mxgrey: Ah, right, this documentation is outdated because of a refactor I did a while back.

rmf_traffic/src/rmf_traffic/schedule/ChangeInternal.hpp
```diff
@@ -0,0 +1,64 @@
...
+
+#include <rmf_traffic/schedule/Change.hpp>
+
+#include <rmf_utils/optional.hpp>
+
+namespace rmf_traffic {
+namespace schedule {
+
+//==============================================================================
+class Change::Delay::Implementation
```
aaronchongth: is there a reason only `Change::Delay::Implementation` is in `ChangeInternal.hpp` while others for `Add`, `Erase`, `RegisterParticipant`, `UnregisterParticipant`, `Cull` are not? 

Im guessing there will be no difference since its not in the interface, but just wondering if it was a design decision.

mxgrey: It was done like that because other source files (specifically `Database.cpp`) use `Change::Delay::Implementation` whereas no other source files had any use for the other `Implementation` classes.

rmf_traffic/src/rmf_traffic/schedule/Timeline.hpp
```diff
@@ -0,0 +1,517 @@
...
+        continue;
+
+      const Entries& timeline = map_it->second;
+      const Time* const lower_time_bound = region.get_lower_time_bound();
+      const Time* const upper_time_bound = region.get_upper_time_bound();
+
+      spacetime_data.lower_time_bound = lower_time_bound;
+      spacetime_data.upper_time_bound = upper_time_bound;
+
+      const auto timeline_begin =
```
Yadunund: I'm guessing this was written before a refactor.
```suggestion
      const auto timeline_begin = get_timeline_begin(timeline. lower_time_bound);
```

rmf_traffic/src/rmf_traffic/schedule/Timeline.hpp
```diff
@@ -0,0 +1,517 @@
...
+    {
+      const std::string& map = region.get_map();
+      const auto map_it = _timelines.find(map);
+      if (map_it == _timelines.end())
+        continue;
+
+      const Entries& timeline = map_it->second;
+      const Time* const lower_time_bound = region.get_lower_time_bound();
+      const Time* const upper_time_bound = region.get_upper_time_bound();
+
```
Yadunund: If the region is time bounded, would it be beneficial to check whether it overlaps in time with the given timeline? My understanding is that `rmf_traffic::internal::detect_conflict()` does check for this but only later on in the `inspector() function. Would inserting
```
auto last_it = --timeline.end();
if (lower_time_bound != nullptr && *lower_time_bound > last_it->first)
  continue;
if (upper_time_bound != nullptr && *upper_time_bound < timeline.begin()->first)
  continue;
```
here increase performance in anyway?

rmf_traffic/include/rmf_traffic/Trajectory.hpp
Yadunund: Agreed!

rmf_traffic/src/rmf_traffic/schedule/Timeline.hpp
mxgrey: I might not be thinking about this hard enough, but my instinct tells me that we will effectively get this by the way we iterate from `timeline_begin` to `timeline_end`. If there is no overlap between the time range of the region and the time range of the timeline, then `timeline_begin == timeline_end` and so none of the buckets will be inspected.

Yadunund: Ah yes you're right none of the buckets will be inspected. But we will still be looping through each space in the region and calling `inspect_entries()`? Guessing that doesn't cost much?

mxgrey: I'm not sure how looping through each space relates to filtering out based on time.

There isn't any bucket system for us to filter based on space, so we're stuck looping through all the regions no matter what; at least until we implement a broadphase partition that accounts for space.

mxgrey: Now I understand. Fixed in d7591ee

PR #82, opened 2020-03-03
Test/modular

Yadunund:
Tests to check version number overflow.

### DIFF
--- a/rmf_traffic/test/unit/schedule/test_Modular.cpp
+++ b/rmf_traffic/test/unit/schedule/test_Modular.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2019 Open Source Robotics Foundation
+ * Copyright (C) 2020 Open Source Robotics Foundation
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


mxgrey:
My IDE is alerting me to unsequenced modification warnings for this test, e.g.:

```
test_Modular.cpp:44:40: warning: unsequenced modification and access to '_max_value'
```

What this basically means is that when you're using the increment `++` and decrement `--` operators on a value that is also accessed elsewhere in a single line, the result may be [undefined behavior](https://stackoverflow.com/a/33743361).

I recommend performing the incrementing and decrementing outside of the `CHECK` statements on their own lines. This will make the test look less elegant, but it should eliminate undefined behavior and probably make the test more clear to human readers as well.

rmf_traffic/test/unit/schedule/test_Modular.cpp
```diff
@@ -0,0 +1,51 @@
+/*
+ * Copyright (C) 2019 Open Source Robotics Foundation
```
mxgrey:
```suggestion
 * Copyright (C) 2020 Open Source Robotics Foundation
```

mxgrey:
I felt it would be good to make the tests more explicit instead of using underscores in variable names to indicate arithmetic operations, so I opened #85.

PR #84, opened 2020-03-10
Migrate fleet adapter to use the schedule participant system

mxgrey:
This PR changes the implementation of the fleet adapters (both read only and full control) to use the new schedule participant system.

In terms of code review there probably isn't much of interest to look over, since it's mostly just migrating from the old schedule API to the new one. But it would be good to get some practical tests of this new system.

I've run some rudimentary tests in the DP2 environment, so it might be worth trying this new system out on the new RMF++ example environments that are being developed.

**NOTE:** The conflict resolution system is not totally finished yet. I know that right now the schedule visualizer will permanently display participants as being in conflict after the first time that they encounter a conflict. I also would not be surprised if deadlocks happen pretty easily. My next task is to finish the new, more rigorous, schedule conflict resolution system.


mxgrey:
The "negative delay" has been confusing me a bit. It's especially confusing if it happens after a robot was waiting at a door, because I would expect a positive delay in that case.

Since the fleet adapter is due for a complete rewrite in the near future my plan is to not worry about it for now. But if you happen to identify a pattern or cause, please do share your insights.

PR #86, opened 2020-03-17
Test/profile

Yadunund:
Tests for new Profile class

### DIFF
--- a/rmf_traffic/test/unit/test_Profile.cpp
+++ b/rmf_traffic/test/unit/test_Profile.cpp
@@ -21,34 +21,156 @@
 
 #include <rmf_utils/catch.hpp>
 
-#include <iostream>
-
 using namespace std::chrono_literals;
 
 SCENARIO("Testing Construction")
 {
+  using Profile = rmf_traffic::Profile;
 
-using Profile = rmf_traffic::Profile;
-using Trajecotry = rmf_traffic::Trajectory;
+  WHEN("Vicinity is not passed into the constructor")
+  {
+    const auto shape_1 = rmf_traffic::geometry::make_final_convex<
+        rmf_traffic::geometry::Circle>(1.0);
+    const auto shape_2 = rmf_traffic::geometry::make_final_convex<
+        rmf_traffic::geometry::Circle>(2.0);
 
-WHEN("When vicinity is not passed into the constructor")
-{
-  const auto shape = rmf_traffic::geometry::make_final_convex<
-      rmf_traffic::geometry::Circle>(1.0);
+    auto profile = Profile{shape_1};
 
-  const auto profile = Profile{shape};
+    auto& footprint = profile.footprint();
+    auto& vicinity = profile.vicinity();
 
-  auto& footprint = profile.footprint();
-  auto& vicinity = profile.vicinity();
+    CHECK((footprint->get_characteristic_length() - 1.0) == Approx(0.0).margin(1e-6));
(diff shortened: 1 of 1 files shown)


rmf_traffic/test/unit/test_Profile.cpp
```diff
@@ -0,0 +1,251 @@
...
+        rmf_traffic::geometry::Circle>(1.0);
+    const auto circle_2 = rmf_traffic::geometry::make_final_convex<
+        rmf_traffic::geometry::Circle>(2.0);
+
+    auto profile = Profile{circle_1};
+
+    auto& footprint = profile.footprint();
+    auto& vicinity = profile.vicinity();
+
+    CHECK((footprint->get_characteristic_length() - 1.0) == Approx(0.0).margin(1e-6));
```
mxgrey: Is there a reason to test the difference instead of just testing the value directly? E.g.:

```
CHECK(footprint->get_characteristic_lenght() == Approx(1.0));
```

In other tests we used distance because it's simpler than checking each component, but that's not relevant here since we're testing a scalar value. And I think we can use the default margin instead of specifying one explicitly.

mxgrey: You can also consider testing the identity of the footprint.

```
CHECK(footprint == circle_1);
```

should be a valid test here.

rmf_traffic/test/unit/test_Profile.cpp
```diff
@@ -0,0 +1,251 @@
...
+    CHECK((vicinity->get_characteristic_length() - 1.0) == Approx(0.0).margin(1e-6));
+
+    // Footprint and vicinity are updated
+    profile.vicinity(circle_2);
+    profile.footprint(circle_2);
```
mxgrey: Maybe we should consider a test where `vicinity` gets set while `footprint` remains the same.

rmf_traffic/test/unit/test_Profile.cpp
Yadunund: Thanks for pointing these out. I have modified the checks to test the values directly. Also included checks to test the identity of footprints and vicinities. But not sure if it makes sense to have both though? I could remove the tests for `characteristic_length` if its redundant.

rmf_traffic/test/unit/test_Profile.cpp
Yadunund: I've updated the tests to include sub-blocks for 1) Only footprint is updated 2) Only vicinity is updated and 3) Both footprint and vicinity are updated.

rmf_traffic/test/unit/test_Profile.cpp
```diff
@@ -0,0 +1,294 @@
...
+    CHECK(vicinity == circle_1);
+    CHECK(footprint->get_characteristic_length() == Approx(1.0));
+    CHECK(vicinity->get_characteristic_length() == Approx(1.0));
+
+    WHEN("Only footprint is updated")
+    {
+      profile.footprint(circle_2);
+      CHECK(footprint == circle_2);
+      CHECK(profile.footprint()->get_characteristic_length() == Approx(2.0));
+      // profile.vicinity() should return footprint
```
mxgrey:
```suggestion
      // profile.vicinity() should return footprint
      CHECK(profile.vicinity() == profile.footprint());
```

Let's make that an explicit test instead of just a comment.

PR #87, opened 2020-03-20
New schedule negotiation system

mxgrey:
This introduces a completely reworked system for negotiating schedule conflicts. The new system was unceremoniously crammed into the old fleet adapter implementation, so many of the same limitations and issues of the old system are still present.

Even if not all issues are resolved yet, I think this new implementation should be an improvement over the previous system. Testing on this branch should focus on verifying that situations which worked on the DP2 implementation are still working on this new implementation.

For rmf_schedule_visualizer, you will need to use the osrf/rmf_schedule_visualizer#38 branch.


mxgrey:
After this gets reviewed and approved, my plan is to merge it into #84 and then merge that whole thing into the `develop` branch. Then we can create a PR from `develop` to `master`.

PR #88, opened 2020-03-20
Feature/multiple deliveries

cnboonhan94:
This PR is targeted at full control `rmf_fleet_adapter`s, to 

1) implement concurrent deliveries by multiple robots
2) fix a possible bug in the [previous implementation](https://github.com/osrf/rmf_core/blob/master/rmf_fleet_adapter/src/full_control/FleetAdapterNode.cpp#L475) where the `_have_delivery_request` flag was not unset after completion of tasks, which prevented the processing of further delivery requests after the first.

### DIFF
--- a/rmf_fleet_adapter/src/full_control/FleetAdapterNode.cpp
+++ b/rmf_fleet_adapter/src/full_control/FleetAdapterNode.cpp
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  *
- */
+*/
 
 #include "FleetAdapterNode.hpp"
 
@@ -37,7 +37,8 @@ namespace rmf_fleet_adapter {
 namespace full_control {
 
 //==============================================================================
-std::shared_ptr<FleetAdapterNode> FleetAdapterNode::make() {
+std::shared_ptr<FleetAdapterNode> FleetAdapterNode::make()
+{
   const auto node = std::shared_ptr<FleetAdapterNode>(new FleetAdapterNode);
 
   node->_perform_deliveries =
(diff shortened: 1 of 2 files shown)


mxgrey:
I'll be completely honest: There was something horribly broken with receiving delivery requests, and I put in that terrible `_have_delivery_request` on purpose to deal with it at the last minute. I'm struggling to remember exactly what the issue was, but I vaguely remember that if multiple delivery requests were issued, then the fleet adapter would break in some awful confusing way (maybe a segfault?).

I decided to punt on it and leave it unsolved since we're working on a complete remake of the fleet adapters, but if there's an interest in getting this to work in the old fleet adapters that were made for DP2, then I would recommend testing this extensively in simulation.

cnboonhan94:
it happens that multiple deliveries is necessary for a simulation world we are working on. so far my limited testing has not yielded any segfaults, but ill be sure to update this thread if i see any.
In the meantime, would it be convenient to leave a legacy branch for this temporary fix in the repo for now?

mxgrey:
I don't mind merging it into master; I just wanted to make sure you guys don't get hurt by that absurd bug that I never managed to track down. If your testing has been turning out okay, then I'm happy to move forward with it.

In case I can't remember the bug correctly, you might also want to try the following tests:
* Spam the same delivery request (i.e. a delivery request with the same task ID) multiple times
* Send multiple delivery requests (i.e. delivery requests with different task IDs) before a delivery is finished
* Do the above with multiple robots in the environment

cnboonhan94:
sure thing. will work on the simulation and carry out these testing as well, and report what I find. Thanks!

cnboonhan94:
Hello there! Sorry for the late response. Ive been testing quite a bit with multiple deliveries, and have not seen any issues as you described. There has been * some * side effects which I'm not sure are directly related to multiple deliveries. nevertheless for many situations, especially conflict free situations, it has been quite ok for me.

mxgrey:
Thanks for the contribution!

PR #92, opened 2020-04-06
Improve fleet adapter

mxgrey:
This PR introduces many comprehensive changes to fleet adapter behavior. In particular, it uses the negotiation system to allow the traffic schedule participants to negotiate with each other in a more robust way. I've been testing it extensively on a particularly difficult version of DP2, and most of the kinks are out. There are a few very specific edge cases that need some more work, but I think overall these changes should prove to be a major improvement.

### DIFF
--- a/rmf_traffic/src/rmf_traffic/schedule/Negotiation.cpp
+++ b/rmf_traffic/src/rmf_traffic/schedule/Negotiation.cpp
@@ -17,9 +17,10 @@
 
 #include <rmf_traffic/schedule/Negotiation.hpp>
 
-#include "Modular.hpp"
 #include "Timeline.hpp"
 #include "ViewerInternal.hpp"
+
+#include <rmf_utils/Modular.hpp>
 
 namespace rmf_traffic {
 namespace schedule {
@@ -124,7 +125,7 @@ public:
   const ParticipantId participant;
   const std::size_t depth;
   rmf_utils::optional<Itinerary> itinerary;
-  Version version = std::numeric_limits<Version>::max();
+  rmf_utils::optional<Version> version = rmf_utils::nullopt;
   bool rejected = false;
   TableMap descendants;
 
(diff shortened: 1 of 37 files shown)


rmf_traffic/src/rmf_traffic/schedule/Negotiation.cpp
```diff
@@ -186,13 +187,17 @@ class Negotiation::Table::Implementation
     const auto it = std::remove_if(unsubmitted.begin(), unsubmitted.end(),
                    [&](const ParticipantId p) { return p == participant; });
     assert(it != unsubmitted.end());
-    unsubmitted.erase(it);
+    unsubmitted.erase(it, unsubmitted.end());
```
aaronchongth, line 190: is there a reason we are erasing all the participants after the owner participant as well? Unless the order of `ParticipantId`s is meaningful in another way?

mxgrey: The order of participants in `unsubmitted` isn't meaningful until line 187. At that point, every element in `submitted` from `it` through `end()` are equal to `participant`, because that's how `std::remove_if` works.

This is called the [Erase-remove Idiom](https://en.wikipedia.org/wiki/Erase%E2%80%93remove_idiom) in C++. It's an efficient way of clearing elements out of a vector that fulfill some criteria. In this case, the criteria is having a value that's equal to `participant`.

In theory the vector `unsubmitted` should be given to us without any duplicate entries, so we could be lazier and just do a find -> erase instead of a remove -> erase, but the remove -> erase pattern makes us more robust to any possible upstream errors.

aaronchongth: thanks for the explanation! never knew about this, will look into using it in the future

mxgrey:
Thanks for documenting these issues!

1. Indeed, I haven't implemented the door sharing yet, because I decided that's too complicated to try to squeeze into the current fleet adapter implementation. Right now any vehicle that's trying to go through a door will refuse to budge until it's all the way through, so this of course creates a deadlock when two vehicles are simultaneously trying to pass through in opposite directions. This will need to be fixed in the next iteration on the fleet adapters.

2. I do believe this is because of the lane merging parameters, although another factor that comes up sometimes is the delay between when a plan is proposed versus when it starts to be executed. I've seen that sometimes result in short-cutting, and that's something that needs to be addressed.

PR #94, opened 2020-04-09
some packages dependencies fixes

marcoag:


rmf_traffic/package.xml
```diff
@@ -15,6 +15,9 @@
 
   <build_depend>libccd-dev</build_depend>
   <build_depend>libfcl-dev</build_depend>
+  <build_depend>eigen</build_depend>
```
mxgrey, line 18:
```suggestion

  <depend>eigen</depend>
```

Looks like Eigen gets used in the public API in a few spots (e.g. [here](https://github.com/osrf/rmf_core/blob/23343206a6bf120430542370e8350c5936def2e3/rmf_traffic/include/rmf_traffic/agv/Graph.hpp#L57)) so we should make this a regular `<depend>`.

PR #96, opened 2020-04-14
RMF Linter

Yadunund:
This PR introduces the `rmf_uncurstify` linter and applies the same on
- `rmf_traffic`
- `rmf_traffic_ros2`
- `rmf_fleet_adapter` 

packages based on the rules specified in `rmf_utils/test/format/rmf_code_style.cfg`

Once the packages are built, running
```colcon test --packages-select rmf_traffic rmf_traffic_ros2 rmf_fleet_adapter```
will run the cmake-integrated `rmf_uncristify` linter on the packages.

To see details of the test results
```cat log/latest_test/<pkg_name>/stdout_stderr.log```

### DIFF
--- a/rmf_fleet_adapter/CMakeLists.txt
+++ b/rmf_fleet_adapter/CMakeLists.txt
@@ -12,6 +12,17 @@ endif()
 
 find_package(ament_cmake REQUIRED)
 
+if(BUILD_TESTING)
+  find_package(rmf_cmake_uncrustify REQUIRED)
+  find_file(uncrustify_config_file NAMES "share/format/rmf_code_style.cfg")
+                
+  rmf_uncrustify(
+    ARGN include src test
+    CONFIG_FILE ${uncrustify_config_file}
+    MAX_LINE_LENGTH 80
+  )
+endif()
+
 set(dep_pkgs
   rclcpp
   rmf_utils
--- a/rmf_fleet_adapter/src/door_supervisor/Node.cpp
+++ b/rmf_fleet_adapter/src/door_supervisor/Node.cpp
@@ -31,21 +31,21 @@ Node::Node()
   const auto default_qos = rclcpp::SystemDefaultsQoS();
 
   _door_request_pub = create_publisher<DoorRequest>(
-        FinalDoorRequestTopicName, default_qos);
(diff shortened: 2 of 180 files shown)


mxgrey:
Thanks for all the work on this, I'm sure it was a slog!

Out of curiosity, is there any tool that could be used to easily generate a patch for the style corrections? Right now it seems that the best we can do is look at the log and snip the patch segments out of that, but it would be nice if we could just produce the raw patch so we can easily apply it to the code. If that doesn't exist yet, we can open a separate ticket for it and move along with merging this.

Yadunund:
Hi Grey,
Running `colcon test` generates an xunit compliant XML file (`uncrustify.xunit.xml`) which summarizes the results of the test and lists the diff of the changes required. This file is located in `build/<pkg_name>/test_results/<pkg_name>/` directory. I'm not sure if there's a way to use this file to directly patch the original code but I'll look into it. I've attached an example file generated with errors for your reference.
[uncrustify.xunit.zip](https://github.com/osrf/rmf_core/files/4476675/uncrustify.xunit.zip)

However, there is a way to automatically apply all the formatting changes in-place and this is what I use after reviewing the diff. With eloquent sourced,
`ament_uncrustify -c src/rmf_core/rmf_utils/test/format/rmf_code_style.cfg <PATH_TO_FILE(s)> --reformat`

luca-della-vedova:
Throwing my 2 cents, I think a "reasonable" way to structure future work and inspect the diff could be:

1. Do your code changes and tests
2. Commit
3. Run the linter with the --reformat
4. Check that everything makes sense (or if you want a patch git diff > linter.patch)
5. git add and commit --amend
6. Push

Does that makes sense?

mxgrey:
@Yadunund thanks, the `ament_uncrustify` tool is almost exactly what I was hoping for. I wonder if we can wrap that tool with something like `rmf_uncrustify` that automatically selects the `rmf_code_style.cfg`. That can be work for a future PR.

@luca-della-vedova I think that's an excellent work flow. I'll definitely be using that myself :ok_hand:

Yadunund:
@mxgrey sure thing, I can look into making `rmf_code_style.cfg` as the default configuration file in a separate PR.

PR #97, opened 2020-04-20
Add blurb about traffic conflict management

mxgrey:
Due to recent inquiries about traffic conflict management, I've added a blurb about it to the FAQ page.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -116,3 +116,56 @@ soon.
 Currently rmf_core requires ROS2 eloquent for certain launch file features. In
 general, we are likely to be using the latest release of ROS2 while doing
 research and development on RoMi-H.
+
+#### How does `rmf_traffic` avoid mobile robot traffic conflicts?
+
+When we are done implementing the traffic management solution, we will be doing a more
+extensive write-up on the conflict avoidance and negotiation methods than what can reasonably
+fit in an FAQ, but here is a quick outline of the methodology. There are two levels to
+traffic deconfliction: (1) prevention, and (2) resolution.
+
+1. **Prevention.** Whenever possible, it would be good to prevent traffic conflicts from
+happening in the first place. To facilitate this, we have implemented a platform agnostic
+[Traffic Schedule Database](https://github.com/osrf/rmf_core/blob/8cad142e5a5f14133e4e865beeac98fd46edb0e7/rmf_traffic/include/rmf_traffic/schedule/Database.hpp). The traffic schedule is a living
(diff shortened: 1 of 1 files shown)


Jason-Zhou1:
HI mxgrey,
I have found in the office_loop of rmf_demos that one robot A has higher priority over another robot B in the conflict resolution. When conflict arises, robot B either wait for the robot A to pass over or choose another route to go to void the conflict. Am I right? Is this negotiation issue solved in the Negotiation.hpp file or other files?
Thank you!

mxgrey:
The Negotiation system is relatively new development. You'd have to check out the `develop` branch (instead of the default `master` branch) of `rmf_core` to see the behavior from the Negotiation system. There are a couple of [known issues](https://github.com/osrf/rmf_core/pull/92#pullrequestreview-391950457) on the `develop` branch that are being resolved right now, which is why it hasn't been merged to master. I expect to have most of those fixed and tested by early next week.

The implementation on the current `master` is a prototype from December. The way it works is when a traffic conflict occurs, the first traffic schedule participant that offers a way to accommodate the conflict will be the one whose accommodation is accepted. Since Robot A and Robot B belong to the same fleet adapter which operates on a single thread, Robot A will always be the first one to compute and transmit an accommodation to the schedule, so Robot A will always be moving out of the way of Robot B. This wasn't intentionally programmed; it's just a coincidental artifact of how the prototype was implemented and how the scenario was configured. The new negotiation system should produce more efficient and more sensible plans.

Jason-Zhou1:
Hi mxgrey,
I have seen some of your slides about the negotiation procedure: [Here are the slides](https://docs.google.com/presentation/d/1zhN_4Wwt64q2Y4fRxK80p07H5AL0rTCt9w60k5b-9Fs/edit?usp=sharing)
The slides describe the negotiation concept and procedure very clearly and easy to understand. Do you have any slides about the conflict detection logic and procedure yet?

mxgrey:
> Do you have any slides about the conflict detection logic and procedure yet?

No, I don't anticipate there will be time for a write-up on this for a long time, probably not for another year. The key idea is that it uses A* while taking time into account, and it performs continuous collision detection for each A* expansion to prune out infeasible branches.

I think the API of the [planner](https://github.com/osrf/rmf_core/blob/master/rmf_traffic/include/rmf_traffic/agv/Planner.hpp) should simple enough to use, though. Just give it a description of your robot, a snapshot of the current schedule, and the start and end conditions. Then it will plan out a conflict-free itinerary for you.

PR #101, opened 2020-05-14
Complete traffic negotiation system

mxgrey:
This doozy of a PR completes the traffic negotiation system and integrates it into the existing fleet adapter implementations. There is also some refactoring for `rmf_traffic::agv::Planner`, `rmf_traffic::schedule::[Database|Mirror]`, and `rmf_traffic::schedule::Negotiation::Table` to facilitate multi-threaded usage.

While this is almost completely backwards compatible with the old fleet adapters, there is one important change: the `profile_radius` parameter of the fleet adapter needs to be split into two new parameters: `footprint_radius` and `vicinity_radius`.

For reviewers: This PR is critical for our June release deadline, so I recommend focusing on just directly testing it against `rmf_demos` and/or any other RMF scenarios that you know of to verify that nothing is broken. We can take deal with performing a deeper code quality check later on when time permits.


mxgrey:
I've now tested out how the `read_only` fleet adapter behaves under negotiations, and it seems to me like it's working correctly now. Let me know if you encounter any issues with it.

Jason-Zhou1:
Hi all,
Who can explain to me how to get the A's ideal proposal in the attached file? Tks.
traffic_negotiation1

mxgrey:
I'm happy to help you find what you're looking for, but I'll need you to elaborate on what you mean by "get" in this context. I think you're probably asking for one of the following, but I'm not sure which:

1. How is A's Ideal Proposal computed?
2. Where in the fleet adapter is A's ideal proposal computed?
3. How can I view what A has computed for its ideal proposal during a live demo?

Depending on which you mean, I would have a very different answer.

Jason-Zhou1:
Hi mxgrey,
I want to know #2. Thank you!

mxgrey:
For that question, the answer goes somewhat deep, and may be a bit convoluted since there are a lot of moving parts involved.

* First, the fleet adapter node will receive a conflict notification which tells it that it needs to participate in this negotiation. This notification gets received by the [`rmf_traffic_ros2::schedule::Negotiation`](https://github.com/osrf/rmf_core/blob/master/rmf_traffic_ros2/include/rmf_traffic_ros2/schedule/Negotiation.hpp) class.
* For each robot, the fleet adapter [subscribes to that object](https://github.com/osrf/rmf_core/blob/master/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp#L323) so that it will be told when the robot needs to respond to a negotiation.
* When a robot needs to respond to a negotiation its [`GoToPlace::Active::respond(~)`](https://github.com/osrf/rmf_core/blob/master/rmf_fleet_adapter/src/rmf_fleet_adapter/phases/GoToPlace.cpp#L88) implementation gets triggered (I'm leaving out some details of how that happens, but I can elaborate if you'd like).
* That function will launch a multi-threaded [`Negotiate`](https://github.com/osrf/rmf_core/blob/master/rmf_fleet_adapter/src/rmf_fleet_adapter/services/Negotiate.hpp) service, whose main implementation can be found [here](https://github.com/osrf/rmf_core/blob/master/rmf_fleet_adapter/src/rmf_fleet_adapter/services/detail/impl_Negotiate.hpp).

Every block in the negotiation diagram is using the `Negotiate` service the exact same way, whether it's `A's Ideal Proposal` or `B's best attempt to accommodate A` or `C's best attempt to accommodate A & B`, etc. The only difference between the various blocks is what constraints they need to deal with. Those constraints are described by the [`rmf_traffic::schedule::Negotiation::Table::Viewer`](https://github.com/osrf/rmf_core/blob/master/rmf_traffic/include/rmf_traffic/schedule/Negotiation.hpp#L143-L208) object that gets passed to the [`respond(~)`](https://github.com/osrf/rmf_core/blob/master/rmf_fleet_adapter/src/rmf_fleet_adapter/phases/GoToPlace.cpp#L89) function. Because the same object can be used to describe the constraints of all the different blocks in the diagram, we can use the same code to solve every block.

You'll probably notice that there are "reject" and "forfeit" code paths that aren't discussed in the above diagram.
* The "rejection" mechanism is used when it's impossible for one of the fleets to accommodate a proposal that came from another. When a rejection is performed, the rejecting fleet will provide a set of feasible trajectories (usually anywhere from 10-200 trajectories) and the fleet that receives the rejection should try once again to find an ideal proposal for itself, but that ideal proposal must accommodate at least one of the trajectory alternatives that were provided with the rejection.
* The "forfeit" mechanism is used when the planner is having an inordinately difficult time finding any kind of solution. This can happen when the negotiation has numerous participants that are all actively on the move, which can lead to situations that are seemingly impossible to resolve due to inconsistencies across time. Typical when a forfeit is used, there will be another feasible combination of accommodations that gets found by the negotiation. In the worst case scenario, if negotiations keep failing, the robots may experience a real-life deadlock. When a deadlock happens, the participants will be sitting still so the negotiation will reach a steady state and will not be negatively affected by async inconsistencies. When that happens, a successful resolution is practically assured.

That's the code path for the "Full Control" style of adapter. The implementation for the "Read Only" and "Traffic Light" APIs are a bit different. I could also cover those if you'd like.

Jason-Zhou1:
Thank you, mxgrey! Now I understand more about the computation of ideal proposal.
But I still cannot understand block relationship in the diagram. I feel that A's ideal proposal should be generated by the Smart Fleet Adapter A, but it is not reflected in the diagram. May you explain a bit more about the question #1 ?
Thanks!

mxgrey:
> I feel that A's ideal proposal should be generated by the Smart Fleet Adapter A, but it is not reflected in the diagram.

A's ideal proposal is indeed generated by the smart fleet adapter A. Perhaps the misunderstanding is that this diagram is one slide out of four. Here is the complete sequence of slides:

slide_1

-------

slide_2

-------

slide_3

-------

slide_4

Jason-Zhou1:
Hi mxgrey,
Would you mind describing what the proposal of a fleet looks like? Thanks!

mxgrey:
Sure, the proposal is fairly simple. It's simply an [itinerary](https://github.com/osrf/rmf_core/blob/4180a944e7497a3dd6b08f17280706ce6aca8a30/rmf_traffic/include/rmf_traffic/schedule/Itinerary.hpp#L33) which is a set of [routes](https://github.com/osrf/rmf_core/blob/4180a944e7497a3dd6b08f17280706ce6aca8a30/rmf_traffic/include/rmf_traffic/Route.hpp) where a route is simply the name of a map (which usually differentiates between different floors in a building) combined with a 2D homogeneous [trajectory](https://github.com/osrf/rmf_core/blob/4180a944e7497a3dd6b08f17280706ce6aca8a30/rmf_traffic/include/rmf_traffic/Trajectory.hpp).

Effectively a proposal is just a trajectory through space-time, described by a sequence of timed waypoints which are interpolated as piecewise cubic splines.

In addition, all negotiation participants always have a [participant description](https://github.com/osrf/rmf_core/blob/4180a944e7497a3dd6b08f17280706ce6aca8a30/rmf_traffic/include/rmf_traffic/schedule/ParticipantDescription.hpp) of all the other participants. The participant description contains a [profile](https://github.com/osrf/rmf_core/blob/4180a944e7497a3dd6b08f17280706ce6aca8a30/rmf_traffic/include/rmf_traffic/Profile.hpp) which describes how much space the participant takes up (i.e. its footprint).

Combining a participant's trajectory with its footprint allows its proposal to be treated as a dynamic obstacle for the other participants to plan around (i.e. produce a trajectory that does not collide with any of the other trajectories). As you get deeper into the negotiation tree, you're effectively adding more of these dynamic obstacles for the negotiator to plan around.

Jason-Zhou1:
Hi grey,
May I know how each fleet manager comes out its proposal? Thanks a lot!

mxgrey:
At the risk of being too pedantic, we make a distinction between "Fleet Manager" and "Fleet Adapter". The "Fleet Manager" is something that usually comes with a robot system and was likely developed before RoMi-H ever existed. The "Fleet Adapter" talks to the Fleet Manager API to issue proposals and handle negotiations on its behalf. If a new Fleet Manager were developed right now, it could interact with RoMi-H directly instead of needing an adapter, but such a fleet manager does not currently exist.

Right now there are three categories of fleet adapter that have explicit APIs: "Full Control", "Traffic Light", and "Read Only".

For "Full Control" and "Traffic Light", the proposals are generated exactly how I described in [this previous comment](https://github.com/osrf/rmf_core/pull/101#issuecomment-677505223). The difference between the "Full Control" and the "Traffic Light" is that the navigation graph for the "Full Control" fleet adapter considers every route that the robot is allowed to follow when it produces a plan, whereas the "Traffic Light" will only consider the robot's current path when planning, and the planner will only be able to make start/stop choices. But all the code to perform the actual negotiation is the same.

The "Read Only" fleet adapter is a bit different. That category of fleet adapter uses the [`StubbornNegotiator`](https://github.com/osrf/rmf_core/blob/master/rmf_traffic/include/rmf_traffic/schedule/StubbornNegotiator.hpp) class. The "Stubborn Negotiator" is a negotiation participant that only accepts a single itinerary for itself. When it makes a proposal, it will always simply propose its current itinerary. When it responds to a proposal from another participant, it will reject that proposal if it conflicts with its own current itinerary. Everything about the negotiation system works exactly the same; it's just that the read-only fleet adapters are stubborn about what they will propose.

Jason-Zhou1:
Hi grey,
I have seen in the "office loop" demo that sometimes one robot does "stop/start" action (traffic light), while othertimes the same robot does "re-planning/detour" action (full control). Are there any rules for the switch between these two control categories? Tks!

mxgrey:
They aren't switching between control categories; the simulations that currently exist are always using the "full control" category.

The planner + negotiation system will choose the optimal overall itinerary. The planner for the "full control" fleet adapter may choose to have a robot stop and wait on a waypoint if doing so would be more optimal than rerouting the robot along a different path.

The "Traffic Light" category is a strictly worse category than "Full Control". The "Full Control" can do anything the "Traffic Light" can do, plus much more. But many fleet managers that exist today can only provide a stop/resume API for externally controlling their robots, making it impossible for them to integrate with a "Full Control" fleet adapter. We provide the "Traffic Light" category for the sole purpose of making RoMi-H accessible to those robot fleets.

Jason-Zhou1:
Oh okay. Thank you Grey for your clarifications!

Jason-Zhou1:
Hi mxgrey,
Sorry I don't know where to put this question about the Romi-H structure. i have found the internal modules in the rmf_core before, as seen in the figure:

image

May I know what are the main contents in the modules “Dispatcher Planner” and "Traffic Monitor"? Or what are the main related cpp files inside respective modules?

Jason-Zhou1:
> For that question, the answer goes somewhat deep, and may be a bit convoluted since there are a lot of moving parts involved.
> 
> * First, the fleet adapter node will receive a conflict notification which tells it that it needs to participate in this negotiation. This notification gets received by the [`rmf_traffic_ros2::schedule::Negotiation`](https://github.com/osrf/rmf_core/blob/master/rmf_traffic_ros2/include/rmf_traffic_ros2/schedule/Negotiation.hpp) class.
> * For each robot, the fleet adapter [subscribes to that object](https://github.com/osrf/rmf_core/blob/master/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp#L323) so that it will be told when the robot needs to respond to a negotiation.
> * When a robot needs to respond to a negotiation its [`GoToPlace::Active::respond(~)`](https://github.com/osrf/rmf_core/blob/master/rmf_fleet_adapter/src/rmf_fleet_adapter/phases/GoToPlace.cpp#L88) implementation gets triggered (I'm leaving out some details of how that happens, but I can elaborate if you'd like).
> * That function will launch a multi-threaded [`Negotiate`](https://github.com/osrf/rmf_core/blob/master/rmf_fleet_adapter/src/rmf_fleet_adapter/services/Negotiate.hpp) service, whose main implementation can be found [here](https://github.com/osrf/rmf_core/blob/master/rmf_fleet_adapter/src/rmf_fleet_adapter/services/detail/impl_Negotiate.hpp).
> 
> Every block in the negotiation diagram is using the `Negotiate` service the exact same way, whether it's `A's Ideal Proposal` or `B's best attempt to accommodate A` or `C's best attempt to accommodate A & B`, etc. The only difference between the various blocks is what constraints they need to deal with. Those constraints are described by the [`rmf_traffic::schedule::Negotiation::Table::Viewer`](https://github.com/osrf/rmf_core/blob/master/rmf_traffic/include/rmf_traffic/schedule/Negotiation.hpp#L143-L208) object that gets passed to the [`respond(~)`](https://github.com/osrf/rmf_core/blob/master/rmf_fleet_adapter/src/rmf_fleet_adapter/phases/GoToPlace.cpp#L89) function. Because the same object can be used to describe the constraints of all the different blocks in the diagram, we can use the same code to solve every block.
> 
> You'll probably notice that there are "reject" and "forfeit" code paths that aren't discussed in the above diagram.
> 
> * The "rejection" mechanism is used when it's impossible for one of the fleets to accommodate a proposal that came from another. When a rejection is performed, the rejecting fleet will provide a set of feasible trajectories (usually anywhere from 10-200 trajectories) and the fleet that receives the rejection should try once again to find an ideal proposal for itself, but that ideal proposal must accommodate at least one of the trajectory alternatives that were provided with the rejection.
> * The "forfeit" mechanism is used when the planner is having an inordinately difficult time finding any kind of solution. This can happen when the negotiation has numerous participants that are all actively on the move, which can lead to situations that are seemingly impossible to resolve due to inconsistencies across time. Typically when a forfeit is used, there will be another feasible combination of accommodations that gets found by the negotiation. In the worst case scenario, if negotiations keep failing, the robots may experience a real-life deadlock. When a deadlock happens, the participants will be sitting still so the negotiation will reach a steady state and will not be negatively affected by async inconsistencies. When that happens, a successful resolution is practically assured.
> 
> That's the code path for the "Full Control" style of adapter. The implementation for the "Read Only" and "Traffic Light" APIs are a bit different. I could also cover those if you'd like.

Hi mxgrey,
So the proposal algorithm mainly realizes in the fleet adapter side, right? What about Negotiation process? I find three places of Negotiation: rmf_traffic_ros2::schedule::Negotiation, rmf_traffic/include/rmf_traffic/schedule/, rmf_fleet_adapter/services/Negotiate.hpp.

mxgrey:
The overall negotiation is a distributed process. Every fleet adapter has its own live copies of the negotiations, and they send messages between each other to keep each other updated on their proposals and rejections. There is also a live copy of each negotiation [in the schedule node](https://github.com/osrf/rmf_core/blob/89e758ac9b4fddc0302dac2d10d062ad7fadb477/rmf_traffic_ros2/src/rmf_traffic_ros2/schedule/internal_Node.hpp#L389-L390). The schedule node watches its copies of each negotiation in order to determine when the negotiation is complete and a conclusion can be announced. Then the schedule node takes care of announcing the conclusion.

Jason-Zhou1:
Hi mxgrey,
Would you please explain more about what are the outputs of proposal? How to determine the complete and conclusion of negotiation?

mxgrey:
[Here are the slides](https://docs.google.com/presentation/d/1zhN_4Wwt64q2Y4fRxK80p07H5AL0rTCt9w60k5b-9Fs/edit?usp=sharing) that I'll be presenting at the ROS-Industrial workshop next week. They go through the concept of the negotiation system in some detail. It's not pseudo-code, but I think it'll be easier to understand than pseudo-code.

The output of a proposal is simply a trajectory for the robot. Specifically a trajectory that accommodates all the trajectories of the parent proposals in the negotiation tree.

A conclusion can be chosen when at least one branch of a negotiation reaches a depth of `N` where `N` is the number of robots. However, this is not guaranteed to be the optimal conclusion. If you wait until all `N!` branches reach a depth of `N`, then you are guaranteed to have the optimal conclusion available. However, `N!` can be quite large if `N` is high, so it might not always be worth waiting until they all arrive. For that reason, the best strategy would probably be to balance out the desire for an optimal conclusion with the desire for a quick response. System integrators are free to decide how to do that for their deployments of RMF.

Jason-Zhou1:
hi mxgrey,
1. Third party judge measures the penalty and determines the best proposal for the fleet. May I know where is the code for this issue in rmf_core if any?
2. I have ever seen some code to describe Itinerary as a class, but cannot always find it in the rmf_core repo. Can you please show me the course code for the complete description Itinerary data type?
3.What's the data type of conclusion? How does the rmf_code code publish it?
Thanks!

mxgrey:
1. The [`QuickestFinishEvaluator`](https://github.com/osrf/rmf_core/blob/master/rmf_traffic/include/rmf_traffic/schedule/Negotiation.hpp#L469) is an example of a third-party judge implementation. The implementation of its selection function can be found [here](https://github.com/osrf/rmf_core/blob/master/rmf_traffic/src/rmf_traffic/schedule/Negotiation.cpp#L1371). Note that this class is implementing the `Negotiation::Evaluator` interface, so all you need to do to make a custom third-party judge is make a custom implementation of that interface.

2. `Itinerary` is an alias for a `std::vector` of `std:: shared_ptr<const Route>` as defined [here](https://github.com/osrf/rmf_core/blob/master/rmf_traffic/include/rmf_traffic/schedule/Itinerary.hpp#L33).

3. The conclusion of a negotiation can be represented as a [`Negotiation::VersionedKeySequence`](https://github.com/osrf/rmf_core/blob/master/rmf_traffic/include/rmf_traffic/schedule/Negotiation.hpp#L112), and gets transmitted using [this message](https://github.com/osrf/rmf_core/blob/master/rmf_traffic_msgs/msg/NegotiationConclusion.msg). The Negotiation is a [trie](https://en.m.wikipedia.org/wiki/Trie) data structure, so you need a sequence of keys to uniquely specify a node in it.

Jason-Zhou1:
Hi mxgrey,
For issue 3, I have found the NegotiationConclusion.msg file. There is an element: bool resolved. I want to know how it is get assigned. Would you please kindly give me an example where it is given value in the negotiation code?
Thanks a lot!

mxgrey:
If enough tables in the negotiation have been forfeited (which is the third option for a negotiator, alongside proposing and rejecting), then a negotiation may be "complete" but unsuccessful. It's "complete" in the sense that the negotiators have given up too many times for a successful negotiation to ever come out of it.

[Here](https://github.com/osrf/rmf_core/blob/master/rmf_traffic_ros2/src/rmf_traffic_ros2/schedule/Node.cpp#L840) is where that case is handled for the ROS2-based schedule node. After the negotiation is declared a failure, the negotiators all give up on it and then be free to enter into a fresh new negotiation. Typically if a negotiation fails it's because of the async nature of the robots trying to move through the environment while simultaneously negotiating for the future. This async property means that something a negotiator proposed earlier becomes impossible to deal with since the real-life situation has changed too much since that proposal. This usually only happens when robots are in very close proximity and experiencing frequent delays outside of their control. In the worst case scenario, the robots will all come to a dead stop at which point the negotiation is no longer asynchronous with reality because reality is essentially frozen. At that point, there should no longer be any reason for a negotiation to fail.

Otherwise if the negotiation is successful, then it will be given a value of [true](https://github.com/osrf/rmf_core/blob/master/rmf_traffic_ros2/src/rmf_traffic_ros2/schedule/Node.cpp#L717) and then all the participants will adhere to the chosen conclusion.

Jason-Zhou1:
Thank you very much mxgrey!

Jason-Zhou1:
Hi mxgrey,
I have found a class in https://github.com/osrf/rmf_core/blob/873cc2eac1afc3aec913d4a84dd295d936bd0526/rmf_traffic/include/rmf_traffic/schedule/Negotiation.hpp#Ln359-L361:
TablePtr table(ParticipantId for_participant, const std::vector<ParticipantId>& to_accommodate);

May i know how the to_accommodate is implemented?

mxgrey:
I don't know how to explain it better than [the existing documentation](https://github.com/osrf/rmf_core/blob/873cc2eac1afc3aec913d4a84dd295d936bd0526/rmf_traffic/include/rmf_traffic/schedule/Negotiation.hpp#L342-L346) already does.

If that's not sufficient, you can look at the source code and see how it gets used. Once you have a more clear understanding of it, we'd be happy to review a pull request that improves the documentation's explanation of the table keys.

With how much help you've received in navigating and understanding the `rmf_core` code, you could consider contributing back documentation to help out others who face the same struggles that you did. Unfortunately our development timeline is too tight for us to spend as much time on documentation as we would like, so community contributions for that are always very welcome.

Jason-Zhou1:
Hi mxgrey,
Thank you for clarification! Ok, I will check through the source codes to understand in details.

Jason-Zhou1:
hi mxgrey,
In pg17 of your previous slides: Here are the slides. You said "WALL-E also tells Baymax about a few different paths that WALL-E would be able to follow". In my understanding, each robot will raise ideal proposal, but here you said a few paths proposed. Can you please give more explanation about it?

mxgrey:
The "few different paths" that WALL-E gives to Baymax on slide 17 are part of its **Rejection** of Baymax's ideal proposal. If there are no Rejections, then only ideal plans are proposed.

This Rejection protocol is to deal with the possibility (as illustrated in the slides) that one robot's ideal proposal is impossible for another robot to accommodate, e.g. the robot would get run over because it can't get out of the way. In the worst case scenario, both robot's ideal proposals might be impossible for the other robots, in which case the Rejection protocol is critical for the negotiation to succeed.

PR #122, opened 2020-06-29
Change deprecated macro to its replacement

gbiggs:
See the release notes here:
https://index.ros.org/doc/ros2/Releases/Release-Foxy-Fitzroy/#ament-export-interfaces-replaced-by-ament-export-targets

PR #123, opened 2020-06-30
Remove deprecated node-name parameters from launch files

gbiggs:
The `node-name` parameter in XML-format launch files has been deprecated and removed in Foxy. Moreover, using both `node-name` and `name` is an error. This PR removes the deprecated `node-name` parameter, leaving the `name` parameter.

These changes are backwards-compatible with Eloquent.


mxgrey:
I would propose waiting on merging this and #122 until after the Release Candidate is shipped to minimize any potential confusion about what should be part of the RC.

gbiggs:
That's fine with me.

Yadunund:

PR #125, opened 2020-07-04
Fix compiler warnings

gbiggs:
This PR:

- removes a few of `std::move` usages that are redundant and preventing RVO, because they are turning a local variable into an rvalue reference.
- removes an unused variable in `rmf_fleet_adaptor/tasks/Delivery.cpp`


Yadunund:

PR #126, opened 2020-07-04
Fix deprecated api usage

gbiggs:
Remove usage of APIs that have been deprecated in Foxy, using their new replacements instead.


Yadunund:

PR #130, opened 2020-07-16
Mitigating memory explosions from difficult negotiations

mxgrey:
This PR should help mitigate the fleet adapter memory explosions that were sometimes being caused by difficult negotiations.

This does not fix the underlying issue of the motion planner utilizing large amounts of memory, but it makes it so that planning work being done for negotiations are managed more carefully, and as a result we should see a much lower ceiling for high the memory explosions can get.

A later PR will address the core issue of the motion planner's memory usage. For now this should PR should be sufficient to prevent fleet adapters from using up so much memory that they risk crashing.

### DIFF
--- a/rmf_fleet_adapter/src/rmf_fleet_adapter/Task.cpp
+++ b/rmf_fleet_adapter/src/rmf_fleet_adapter/Task.cpp
@@ -22,6 +22,8 @@
 
 #include <rxcpp/rx-observable.hpp>
 
+#include <malloc.h>
+
 namespace rmf_fleet_adapter {
 
 //==============================================================================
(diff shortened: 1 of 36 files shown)


rmf_fleet_adapter/src/rmf_fleet_adapter/Task.cpp
```diff
@@ -101,6 +103,16 @@ void Task::_start_next_phase()
     _active_phase_subscription.get().unsubscribe();
     _status_publisher.get_subscriber().on_completed();
 
+    // Sometimes difficult negotiations end up seizing an exceedingly large
+    // amount of RAM. This function is used to occasionally the operating system
```
gbiggs, line 107:
```suggestion
    // amount of RAM. This function is used to allow the operating system
```

gbiggs:
Looks like this PR is more than just the trim call, so I don't feel qualified to approve it. I did have one really tiny suggested change though.

mxgrey:
Yeah, the trim is really just to make sure that the explosions are truly temporary and to not alarm anyone who looks at the memory usage of the fleet adapter. The other changes are what was really needed to keep the memory explosions from being unsustainably large, in particular these changes are making it so that difficult planning jobs are being discarded when they're no longer useful instead of being allowed to grow obscenely large and eventually die off with no benefit.

PR #132, opened 2020-07-20
Add negotiation status messages and publishing

ddengster:
As per title

### DIFF
--- a/rmf_traffic_ros2/include/rmf_traffic_ros2/StandardNames.hpp
+++ b/rmf_traffic_ros2/include/rmf_traffic_ros2/StandardNames.hpp
@@ -53,6 +53,8 @@ const std::string NegotiationForfeitTopicName = Prefix +
   "negotiation_forfeit";
 const std::string NegotiationConclusionTopicName = Prefix +
   "negotiation_conclusion";
+const std::string NegotiationStatusTopicName =  Prefix +
+  "negotiation_status";
 
 const std::string EmergencyTopicName = "fire_alarm_trigger";
 
--- a/rmf_traffic_msgs/CMakeLists.txt
+++ b/rmf_traffic_msgs/CMakeLists.txt
@@ -45,6 +45,8 @@ set(msg_files
   "msg/NegotiationRefusal.msg"
   "msg/NegotiationRejection.msg"
   "msg/NegotiationRepeat.msg"
+  "msg/NegotiationStatus.msg"
+  "msg/NegotiationStatusTable.msg"
   "msg/ScheduleInconsistency.msg"
   "msg/ScheduleInconsistencyRange.msg"
   "msg/ScheduleParticipantPatch.msg"
--- a/rmf_traffic_msgs/msg/NegotiationStatus.msg
+++ b/rmf_traffic_msgs/msg/NegotiationStatus.msg
@@ -1,0 +1,10 @@
+
+# The version number assigned to this conflict
+uint64 conflict_version
+
+# The IDs of the participants that are in conflict.
+uint64[] participants
+
+# Array of all tables
+NegotiationStatusTable[] tables
+
--- a/rmf_traffic_msgs/msg/NegotiationStatusTable.msg
+++ b/rmf_traffic_msgs/msg/NegotiationStatusTable.msg
@@ -18,8 +18,18 @@ bool rejected
 # Ongoing status
 bool ongoing
 
+# Defunct status
+bool defunct
+
 # depth in tree of nodes used for sorting
 int32 depth
 
-# index to parent in the NegotiationStatus message. -1 indicates no parent
+# index to tables array in the NegotiationStatus message. -1 indicates no parent
 int32 parent_index
+
+# participant id of proposals. same count as proposals
+uint64[]  proposals_id
+
+# proposals. same count as proposals_id
+Itinerary[] proposals
+
--- a/rmf_traffic_msgs/msg/NegotiationStatusTable.msg
+++ b/rmf_traffic_msgs/msg/NegotiationStatusTable.msg
@@ -1,0 +1,25 @@
+
+# The sequence key that refers to this table. This is equivalent to
+# [to_accommodate..., for_participant]
+uint64[] sequence
+
+# indices to table array of NegotiationStatus.msg
+uint32[] children
+
+# Finished status
+bool finished
+
+# Forfeit status
+bool forfeited
+
+# Rejected status
+bool rejected
+
+# Ongoing status
+bool ongoing
+
+# depth in tree of nodes used for sorting
+int32 depth
+
+# index to parent in the NegotiationStatus message. -1 indicates no parent
+int32 parent_index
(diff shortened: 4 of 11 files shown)


mxgrey:
This PR would duplicate all of the data that's being transmitted about negotiations many many times over. For large negotiations this could be an enormous amount of data being transmitted for every small update.

I understand that dealing with a big chunk of data makes some things easier, but we should be very conscientious of scaling here.

I think what we should do instead is expand on the API of the [`rmf_traffic_ros2::schedule::Negotiation`](https://github.com/osrf/rmf_core/blob/master/rmf_traffic_ros2/include/rmf_traffic_ros2/schedule/Negotiation.hpp) class. That class takes care of aggregating and validating the async negotiation messages. Right now it's only designed to talk with "negotiators" which are objects that can respond to a negotiation, but we could view/snapshot API where you can get an immutable view of the current negotiations that are active.

Let me know if that sounds like an acceptable alternative to this PR.

rmf_traffic_ros2/include/rmf_traffic_ros2/StandardNames.hpp
```diff
@@ -53,6 +53,8 @@ const std::string NegotiationForfeitTopicName = Prefix +
   "negotiation_forfeit";
 const std::string NegotiationConclusionTopicName = Prefix +
   "negotiation_conclusion";
+const std::string NegotiationStatusTopicName =  Prefix +
```
gbiggs, line 56: Is this prefix added to all topics published by this node? If so...

I can see this can't be changed now because all the lines above to the same thing, but putting a prefix on every topic is not a good idea. If some have the prefix and some don't, that's fine, but that doesn't appear to be the case here. It is better to let the launch system add a namespace to all the topics. Hard-coding the prefix circumvents the launch system, reducing flexibility.

mxgrey: The prefixing was my idea, about a year ago. The motivation was to make it explicit that these topics are internal to the traffic management/negotiation system and probably shouldn't be touched by anything outside of that.

If we wanted to further namespace these topics (e.g. to have two independent traffic management/negotiation systems on a single network) would that be something that can still be done with the launch system, or does explicitly having this namespace somehow prohibit that?

ddengster:
> I think what we should do instead is expand on the API of the rmf_traffic_ros2::schedule::Negotiation class. That class takes care of aggregating and validating the async negotiation messages. Right now it's only designed to talk with "negotiators" which are objects that can respond to a negotiation, but we could add a view/snapshot API where you can get an immutable view of the current negotiations that are active.

The aggregation and snapshot api sounds good as a future enhancement, but I'm not sure if that's meant to be the purpose of the rmf_traffic_ros2::schedule::Negotiation class? From the short amount of time I looked at it, it seems to be a response based system. I'd still have to make/find that view/snapshot api.

Once we have that view/snapshot api, we can either call a function every few milliseconds to get a snapshot of the tables, then publish it, or use the negotiators.

I'm guessing here, if we have <4 robots in a conflict at any one time, we should have < 100 tables so it should still be fine when deployed? But anyway, my priority has been getting that romi-dashboard negotiation panel up and listening to a topic. I'd need to cherry pick the negotiation status messages at least in another PR.

mxgrey:
> From the short amount of time I looked at it, it seems to be a response based system.

Its role is to aggregate negotiation data and ensure consistency across the distributed negotiation participants. It just happens to be the case that the only nodes using it so far are negotiators who need to respond to negotiation updates.

> Once we have that view/snapshot api, we can either call a function every few milliseconds to get a snapshot of the tables, then publish it, or use the negotiators.

I might suggest designing it so that it triggers a callback instead of polling it, but I don't know enough about the pipeline into the dashboard to really judge. At the very least if the polling/publishing can be done outside of the Schedule Node that would be much much better, because it's critical that the Schedule Node should not get delayed for the sake of publishing duplicated data.

> I'm guessing here, if we have <4 robots in a conflict at any one time, we should have < 100 tables so it should still be fine when deployed?

We're targeting much much larger systems in the long run, and rejections can make the negotiation data grow considerably.

> my priority has been getting that romi-dashboard negotiation panel up and listening to a topic.

I think this is fine as a proof of concept, but the Schedule Node is the single most critical component in the traffic negotiation system, and burdening it with these publications could have dire effects on the overall system if the Schedule Node becomes a bottleneck.

I don't think these changes should be merged to master until any potential overhead of the negotiation viewing can be isolated from the critical path of the negotiation system. Expanding on the API of `rmf_traffic_ros2::schedule::Negotiation` should be an elegant way of doing that.

If you go with a polling approach, I might suggest adding a function like:

```
std::map<std::size_t, std::vector<rmf_traffic::schedule::Negotiation::Table::ViewerPtr> get_statuses() const;
```

The return value would map from the conflict version to a vector of terminal table viewers. You just need to make sure that the polling function doesn't get called while any negotiations are being modified, or else you'll get undefined behavior.

If you go with a callback approach, I might recommend adding a member function like:

```
void on_status_update(
  std::function<void(std::size_t conflict_version, rmf_traffic::schedule::Negotiation::Table::ViewerPtr)> table_status,
  std::function<void(std::size_t conflict_version, rmf_traffic::schedule::Negotiation::VersionedKeySequence)> concluded,
  std::function<void(std::size_t)> abandoned);
```

Then every time a negotiation is updated, the `table_status` callback will be triggered with a view of the terminal table's new status. `concluded` would be triggered if a negotiation was successfully concluded, and `abandoned` would be triggered if a negotiation had to be forfeited.

In either case, we'll want to add the following member functions to `rmf_traffic::schedule::Negotiation::Table::Viewer`:

```
bool rejected() const;
bool forfeited() const;
```

Adding those should be a simple matter of copying the boolean values into the `Viewer` object when it gets initialized. It's important that we pass back `ViewerPtr`s instead of `ConstTablePtr`s to make sure that we don't get any simultaneous read/write errors.

Let me know if you have any questions about what I've described.

rmf_traffic_ros2/include/rmf_traffic_ros2/StandardNames.hpp
gbiggs: > make it explicit that these topics are internal to the traffic management/negotiation system

That's a sensible use of a prefix. I don't think the current name makes that obvious, though.

> If we wanted to further namespace these topics (e.g. to have two independent traffic management/negotiation systems on a single network) would that be something that can still be done with the launch system, or does explicitly having this namespace somehow prohibit that?

You can still add namespaces using the launch system. The topic names, even with the prefix, are still relative. Starting hard-coded topic names with a slash is the biggest no-no because that completely blocks remapping. Having a prefix just restricts it, for better or worse depending on the use case.

mxgrey:
Oh, one detail I overlooked in the APIs that I proposed is they should have a way to report which participants are involved in each negotiation. Maybe we should have something like this instead..

For polling:
```
struct NegotiationState
{
  std::vector<rmf_traffic::schedule::ParticipantId> participants;
  std::vector<rmf_traffic::schedule::Negotiation::Table::ViewerPtr> tables;
};

std::map<std::size_t, NegotiationState> get_states() const;
```

For reacting:
```
void on_status_update(
  std::function<void(std::size_t conflict_version, std::vector<rmf_traffic::schedule::ParticipantId)> new_negotiation,
  std::function<void(std::size_t conflict_version, rmf_traffic::schedule::Negotiation::Table::ViewerPtr)> table_status,
  std::function<void(std::size_t conflict_version, rmf_traffic::schedule::Negotiation::VersionedKeySequence)> concluded,
  std::function<void(std::size_t)> abandoned);
```

PR #136, opened 2020-07-23
Improve conflict detection for vehicles that start close to each other

mxgrey:
When two vehicles start too close to each other, we need to tweak the definition of what is considered a conflict. Ordinarily a conflict would happen any time the footprint of one robot enters the vicinity of another. However, if a footprint of one is already inside the vicinity of another (which can happen due to latency, localization errors, etc), then it would be impossible to resolve the conflict since there would be no choice for either vehicle that would be considered conflict-free.

Therefore, when vehicles start in close proximity, we need to use a different definition of what constitutes a conflict. With this PR, when two itineraries start in close proximity, it will only be a conflict if either:
1. The derivative of the distance between the two itineraries is ever negative while they are still in close proximity, or
2. They re-invade each other's proximity after exiting.

This change should help to fix deadlocks properly in close-encounter situations.

### DIFF
--- a/rmf_fleet_adapter/CHANGELOG.rst
+++ b/rmf_fleet_adapter/CHANGELOG.rst
@@ -5,7 +5,7 @@ Changelog for package rmf_fleet_adapter
 Forthcoming
 -----------
 
-1.0.1 (2020-07-XX)
+1.0.1 (2020-07-20)
 ------------------
 * Interrupt dangling negotiation planning efforts to reduce memory usage: [#130](https://github.com/osrf/rmf_core/pull/130/)
 * Trim the amount of system memory that is committed to a fleet adapter after each task: [#130](https://github.com/osrf/rmf_core/pull/130/)
--- a/rmf_traffic/CHANGELOG.rst
+++ b/rmf_traffic/CHANGELOG.rst
@@ -5,7 +5,11 @@ Changelog for package rmf_traffic
 Forthcoming
 -----------
 
-1.0.1 (2020-07-XX)
+1.0.2 (2020-07-27)
+------------------
+* Improved definition of "traffic conflict" for vechiles that start too close: [#136](https://github.com/osrf/rmf_core/pull/136)
+
+1.0.1 (2020-07-20)
 ------------------
 * Allow users to specify a callback for interrupting a planner: [#130](https://github.com/osrf/rmf_core/pull/130/)
 * Allow a Negotiation Table Viewer to know when its Table is defunct: [#130](https://github.com/osrf/rmf_core/pull/130/)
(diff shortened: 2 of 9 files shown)


ddengster:
I'm gettting a few test failures, might be related to changes in DetectConflict

```
ddeng@ddeng-PC19:~/is_ws$ ./build/rmf_traffic/test_rmf_traffic 

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
test_rmf_traffic is a Catch v2.5.0 host application.
Run with -? for options

-------------------------------------------------------------------------------
Scenario: A Single Lane
      Given: 2 Participants
       When: Schedule:[], Negotiation:[p0(A->B), p1(A->C)]
       Then: No Proposal is found
-------------------------------------------------------------------------------
/home/ddeng/is_ws/src/rmf_core/rmf_traffic/test/unit/agv/test_Negotiator.cpp:734
...............................................................................

/home/ddeng/is_ws/src/rmf_core/rmf_traffic/test/unit/agv/test_Negotiator.cpp:737: FAILED:
  REQUIRE_FALSE( proposal )
with expansion:
  !{?}




-------------------------------------------------------------------------------
Scenario: DetectConflict unit tests
      Given: A 2-point trajectory t1 with unit square box profile (stationary
             robot)
       When: t1 and t2 are identical and overlapping
       Then: The between function reports the details of conflict
-------------------------------------------------------------------------------
/home/ddeng/is_ws/src/rmf_core/rmf_traffic/test/unit/test_Conflict.cpp:58
...............................................................................

/home/ddeng/is_ws/src/rmf_core/rmf_traffic/test/unit/test_Conflict.cpp:62: FAILED:
  CHECK( conflicts.size() == 1 )
with expansion:
  0 == 1

-------------------------------------------------------------------------------
Scenario: DetectConflict unit tests
      Given: A 2-point trajectory t1 with unit square box profile (stationary
             robot)
       When: t2 is t1 but displaced positionally, with bare overlap at the
             profile within error margin
       Then: The between function reports the details of conflict
-------------------------------------------------------------------------------
/home/ddeng/is_ws/src/rmf_core/rmf_traffic/test/unit/test_Conflict.cpp:87
...............................................................................

/home/ddeng/is_ws/src/rmf_core/rmf_traffic/test/unit/test_Conflict.cpp:90: FAILED:
  CHECK( conflicts.size() == 1 )
with expansion:
  0 == 1

===============================================================================
test cases:    55 |    53 passed | 2 failed
assertions: 11253 | 11250 passed | 3 failed

```

mxgrey:
Good catch, @ddengster! I must have neglected to run those tests. The tests should be revised in the latest commit.

ddengster:
Tests are passing, LGTM!

PR #137, opened 2020-07-23
Bump to version 1.0.2

mxgrey:
There are some more patch fixes upcoming that would be good to push out before the next minor release.


marcoag:
Pushing changes to the buildfarm right now.

PR #138, opened 2020-07-24
Always respond to negotiations

mxgrey:
In some cases race conditions can cause a negotiation to get "lost" for a participant, in particular if a negotiation is still ongoing while a participant finishes its current task.

This PR fixes that problem so that if a negotiation `Responder` dies or a certain amount of time passes, then the negotiation is automatically forfeited. This guarantees that the negotiation never gets stuck and that the schedule never gets locked up trying to wait for a negotiation to finish.


ddengster:
LGTM

PR #140, opened 2020-07-24
[ready] Negotiation messages and callbacks

ddengster:
- Callbacks (StatusUpdateCallback, StatusConclusionCallback) for negotiation status updates and conclusions have been added

- Table::Views now have reject and forfeit flags


rmf_traffic_ros2/include/rmf_traffic_ros2/schedule/Negotiation.hpp
```diff
@@ -68,6 +68,43 @@ class Negotiation
...
+  /// \param[in] cb
+  ///   The callback function to be called upon status conclusions.
+  void on_conclusion(StatusConclusionCallback cb);
+
+  /// Get a Negotiation::TableView that provides a view into what participants are
+  /// proposing.
+  ///
+  /// This function does not care about table versioning.
+  ///
+  /// \sa find()
```
mxgrey, line 97: I think this `\sa find()` was probably copy/pasted from the `rmf_traffic::schedule::Negotiation` documentation, because there doesn't seem to be a `find(~)` function in this class.

rmf_traffic_ros2/src/rmf_traffic_ros2/schedule/Negotiation.cpp
```diff
@@ -858,8 +888,14 @@ class Negotiation::Implementation
...
     }
 
+    if (conclusion_callback)
+      conclusion_callback(msg.conflict_version, msg.resolved);
+    
     // Erase these entries because the negotiation has concluded
-    negotiations.erase(negotiate_it);
+    // NOTE(ddengster): retained_history_count is used for testing 
+    // and doesnt have full functionality
+    if (negotiations.size() > retained_history_count)
```
mxgrey, line 897: This implementation won't really do what the API suggests. It will save the first `N` negotiations that ever occur, and then discard all the rest as they finish.

Instead of this, I would recommend adding a field to the `Implementation`:

```
std::map<Version, rmf_traffic::schedule::Negotiation> history;
```

to store the finished negotiations. Then here you can do:

```
if (retained_history_count > 0)
{
  while (history.size() >= retained_history_count)
    history.erase(history.begin());

  history.insert({msg.conflict_version, std::move(negotiate_it->room.negotiation)});
}
```

The `std::move` semantics will be important here because the `rmf_traffic::schedule::Negotiation` class cannot be copied.

After this, you can tweak the implementation of `table_view(~)` so that it checks in `history` if the negotiation can't be found in `negotiations`.

rmf_traffic_ros2/src/rmf_traffic_ros2/schedule/Negotiation.cpp
```diff
@@ -949,6 +985,45 @@ class Negotiation::Implementation
...
+  TableViewPtr table_view(
+    uint64_t conflict_version,
+    const std::vector<ParticipantId>& sequence)
+  {
+    const auto negotiate_it = negotiations.find(conflict_version);
+    if (negotiate_it == negotiations.end())
+    {
+      RCLCPP_WARN(node.get_logger(), "Conflict version %llu does not exist."
+        "It may have been successful and wiped", conflict_version);
+      return TableViewPtr();
```
mxgrey, line 1003: Nitpick: You can do `return nullptr;` here. I find that to be more explicit that a null value is being returned.

rmf_traffic_ros2/src/rmf_traffic_ros2/schedule/Negotiation.cpp
```diff
@@ -949,6 +985,45 @@ class Negotiation::Implementation
...
+    }
+
+    auto& room = negotiate_it->second.room;
+    Negotiation& negotiation = room.negotiation;
+    auto table = negotiation.table(sequence);
+    if (!table)
+    {
+      RCLCPP_WARN(node.get_logger(), "Table not found");
+      return TableViewPtr();
```
mxgrey, line 1012: Nitpick: You can do `return nullptr;` here.

mxgrey (changes requested):
There are some extremely valuable new features in this PR, but there are a few implementation issues that need to be resolved before we can merge. Details are given below.

rmf_traffic/src/rmf_traffic/schedule/Negotiation.cpp
```diff
@@ -145,6 +145,9 @@ class Negotiation::Table::Viewer::Implementation
   rmf_utils::optional<ParticipantId> parent_id;
   VersionedKeySequence sequence;
   std::shared_ptr<const bool> defunct;
+  std::weak_ptr<const bool> rejected;
```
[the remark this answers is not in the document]
ddengster: switched to bool e9b7db4

rmf_traffic/src/rmf_traffic/schedule/Negotiation.cpp
```diff
@@ -237,8 +240,8 @@ class Negotiation::Table::Implementation
   const ParticipantId participant;
   const std::size_t depth;
   rmf_utils::optional<Itinerary> itinerary;
-  bool rejected = false;
-  bool forfeited = false;
+  std::shared_ptr<bool> rejected = std::make_shared<bool>(false);
```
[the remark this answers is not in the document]
ddengster: switched to bool e9b7db4

rmf_traffic/src/rmf_traffic/schedule/Negotiation.cpp
```diff
@@ -1043,6 +1046,24 @@ bool Negotiation::Table::Viewer::defunct() const
...
 }
 
+bool Negotiation::Table::Viewer::rejected() const
+{
+  return _pimpl->rejected.lock().get();
+}
+
+bool Negotiation::Table::Viewer::forfeited() const
+{
+  return _pimpl->forfeited.lock().get();
```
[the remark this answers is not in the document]
ddengster: switched to bool e9b7db4

rmf_traffic/src/rmf_traffic/schedule/Negotiation.cpp
```diff
@@ -1138,7 +1162,7 @@ bool Negotiation::Table::reject(
 //==============================================================================
 bool Negotiation::Table::rejected() const
 {
-  return _pimpl->rejected;
+  return _pimpl->rejected.get();
```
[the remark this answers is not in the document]
ddengster: switched to bool e9b7db4

rmf_traffic/src/rmf_traffic/schedule/Negotiation.cpp
```diff
@@ -1150,7 +1174,7 @@ void Negotiation::Table::forfeit(Version version)
 //==============================================================================
 bool Negotiation::Table::forfeited() const
 {
-  return _pimpl->forfeited;
+  return _pimpl->forfeited.get();
```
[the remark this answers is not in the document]
ddengster: switched to bool e9b7db4

rmf_traffic_ros2/src/rmf_traffic_ros2/schedule/Negotiation.cpp
ddengster: 94509f9

rmf_traffic_ros2/src/rmf_traffic_ros2/schedule/Negotiation.cpp
ddengster: fixed 94509f9

rmf_traffic_ros2/include/rmf_traffic_ros2/schedule/Negotiation.hpp
ddengster: fixed documentation 94509f9

rmf_traffic/src/rmf_traffic/schedule/Negotiation.cpp
```diff
@@ -1043,6 +1046,24 @@ bool Negotiation::Table::Viewer::defunct() const
   return *_pimpl->defunct;
 }
```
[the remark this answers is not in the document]
ddengster: fixed style 94509f9

rmf_traffic_msgs/msg/NegotiationStatus.msg
```diff
@@ -0,0 +1,23 @@
+
+# The version number assigned to this conflict
```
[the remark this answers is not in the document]
ddengster: these messages and topics are used by `romi-dashboard` and the `rmf_schedule_visualizer`'s websocket server

Also this dependency for `romi-dashboard`, https://github.com/osrf/romi-js-core-interfaces pulls messages from rmf_core. We could put it into rmf_schedule_visualizer but that would create more dependencies for `romi-js-core-interfaces`, so I'd rather it stay in `rmf_core` for now.

rmf_traffic_msgs/msg/NegotiationStatusConclusion.msg
```diff
@@ -0,0 +1,6 @@
+
+# The version number assigned to this conflict
```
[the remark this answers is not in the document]
ddengster: see https://github.com/osrf/rmf_core/pull/140/files#discussion_r472726662

rmf_traffic_ros2/include/rmf_traffic_ros2/StandardNames.hpp
```diff
@@ -53,6 +53,10 @@ const std::string NegotiationForfeitTopicName = Prefix +
   "negotiation_forfeit";
 const std::string NegotiationConclusionTopicName = Prefix +
   "negotiation_conclusion";
+const std::string NegotiationStatusTopicName =
```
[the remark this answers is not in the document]
ddengster: see https://github.com/osrf/rmf_core/pull/140/files#discussion_r472726662

rmf_traffic/include/rmf_traffic/schedule/Negotiation.hpp
```diff
@@ -192,6 +192,11 @@ class Negotiation
       /// remain in sync with the state of the source Table.
       bool defunct() const;
 
+      bool rejected() const;
```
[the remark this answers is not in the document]
ddengster: add documentation 8c2223f

rmf_traffic/src/rmf_traffic/schedule/Negotiation.cpp
```diff
@@ -1043,6 +1046,24 @@ bool Negotiation::Table::Viewer::defunct() const
...
 
+bool Negotiation::Table::Viewer::rejected() const
+{
+  return _pimpl->rejected.lock().get();
```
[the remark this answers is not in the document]
ddengster: oh i missed that, my bad

rmf_traffic_ros2/src/rmf_traffic_ros2/schedule/Negotiation.cpp
ddengster: yea I put it there as an aid for testing, didn't fully implement it, hence the NOTE (it should really be a TODO)

also, i believe std::map does not preserve the order of insertion? std::deque would be a better alternative

rmf_traffic_msgs/msg/NegotiationStatus.msg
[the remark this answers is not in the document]
mxgrey: This hasn't been documented explicitly yet, but the `rmf_traffic_msgs` package is meant for internal use by `rmf_traffic_ros2`. I don't want to commit to API or ABI stability for external use of these message definitions.

It looks like the only message definitions from `rmf_traffic_msgs` that are used in `romi-js-core-interfaces` are these two. I think that's a strong hint that these message definitions may belong elsewhere.

In particular, it's not clear to me why we would want to pack this information to a ROS2 message just to send it across SOSS and be translated into a websocket message when there aren't any other ROS2 participants subscribing to these messages. The usual motivation for doing a ROS2 --> SOSS --> Websocket translation is so that a websocket client tap into publications on a ROS2 system.

Since these messages won't have any other ROS2 listeners, I think it would make more sense to send them directly over a websocket server instead of having an unnecessary translation in the middle. At that point we won't need these message definitions anymore.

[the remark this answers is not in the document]
ddengster: It used to be the case whereby these messages were used by the schedule visualizer for rviz. Likewise, there was existing code being used by `romi-dashboard` to listen to ROS2 messages.

But anyway, the whole thing flows like this: 
ROS2 NegotiationStatus/Conclusion messages go to `romi-dashboard` from the server to construct/render the TreeView ui. When the user clicks on a Conflict/Table UI it requests the negotiation trajectory from the server which the server provides and gives to the dashboard upon receiving.

Some benefits of this approach are that you don't have to setup a whole new protocol for status updates, leaving me to only define the protocol for negotiation trajectories. The other is that you don't have to store as many trajectories on the frontend.

Also there is the issue that the websocket server and `romi-dashboard` was designed as a response to request model (request from romi-dashboard to websocket server), meaning there's significantly more setup to periodically listen for status update from `romi-dashboard`

rmf_traffic_ros2/src/rmf_traffic_ros2/schedule/Negotiation.cpp
mxgrey: > std::deque would be a better alternative.

I considered `std::deque`, but then looking up a specific negotiation has O(N) complexity.

> i believe std::map does not preserve the order of insertion? 

`std::map` does preserve the order of the `Version` value, which will reflect the order that the negotiations were created in. So the history would contain the last `N` completed negotiations based on their time of creation rather than their time of completion. I think that's a reasonable compromise between complexity and intent.

> Mind if I put it these changes into a different PR?

I think the changes should be straightforward enough to get them into this PR. It would just be the code I put in my last comment plus maybe two more lines in `table_view(~)`. We're expected to maintain a high level of quality control for `rmf_core`, so I'd be reluctant to merge a feature into `master` that doesn't do what it describes.

ddengster: > I considered std::deque, but then looking up a specific negotiation would have O(N) complexity.

This is a feature to aid testing, and I always intended N to be small(<10). But it's fine, we can switch to map.

rmf_traffic_msgs/msg/NegotiationStatus.msg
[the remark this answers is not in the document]
mxgrey: > It used to be the case whereby these messages were used by the schedule visualizer for rviz.

The rviz tool is primarily for internal development use as well, so we can afford some flexibility on that. We don't expect any operators to be using rviz in production.

> there is the issue that the websocket server and romi-dashboard was designed as a response to request model (request from romi-dashboard to websocket server), meaning there's significantly more setup to periodically listen for status update from romi-dashboard

The suggestion I'm making is to use [the existing schedule visualization websocket server](https://github.com/osrf/rmf_schedule_visualizer/blob/feature/negotiation-visualizer/rmf_schedule_visualizer/src/Server.hpp) to push these status updates instead of publishing them to a ROS2 topic and then having SOSS translate that and send it over another websocket server.

The websocket client wouldn't periodically poll for status updates; you just need to have the websocket client listen for them. This would eliminate an unnecessary ROS2 --> SOSS --> Websocket translation; we wouldn't need these message definitions anymore; and we wouldn't need any changes to `romi-js-core-interfaces`.

Either way the client needs to listen for the status updates from a websocket server, it's just that the current way requires multiple translation steps and two static message definitions that need to be compiled, while what I'm suggesting would be a more direct websocket communication and would involve two relatively simple JSON message definitions.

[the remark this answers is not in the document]
mxgrey: I guess the gist of what I'm recommending is to merge the [`NegotiationStatusPublisher`](https://github.com/osrf/rmf_schedule_visualizer/blob/feature/negotiation-visualizer/rmf_schedule_visualizer/src/NegotiationStatusPublisher.hpp) with the [`Server`](https://github.com/osrf/rmf_schedule_visualizer/blob/feature/negotiation-visualizer/rmf_schedule_visualizer/src/Server.hpp) and do all the message passing over the visualization websocket server. I don't see any particular reason for the message passing to be fragmented the way it currently is.

rmf_traffic_ros2/src/rmf_traffic_ros2/schedule/Negotiation.cpp
ddengster: reworked e3f0902

rmf_traffic_msgs/msg/NegotiationStatus.msg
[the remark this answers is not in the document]
ddengster: removed a4d34ef

rmf_traffic_msgs/msg/NegotiationStatusConclusion.msg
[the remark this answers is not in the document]
ddengster: remvoed a4d34ef

rmf_traffic_ros2/include/rmf_traffic_ros2/StandardNames.hpp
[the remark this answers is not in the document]
ddengster: removed a4d34ef

rmf_traffic/src/rmf_traffic/schedule/Negotiation.cpp
```diff
@@ -1296,7 +1323,7 @@ auto Negotiation::evaluate(const Evaluator& evaluator) const -> ConstTablePtr
 
     const auto& proposal = Table::Implementation::get(*table_ptr).proposal;
     assert(Table::Implementation::get(*table_ptr).itinerary);
-    assert(!Table::Implementation::get(*table_ptr).rejected);
+    assert(!*Table::Implementation::get(*table_ptr).rejected);
```
mxgrey, line 1326: I think maybe you haven't tried compiling the latest changes in debug mode, because this should produce a compilation error when `NDEBUG` is not defined.

rmf_traffic_ros2/include/rmf_traffic_ros2/schedule/Negotiation.hpp
```diff
@@ -68,6 +68,48 @@ class Negotiation
...
+  ///
+  /// This function does not care about table versioning.
+  /// \param[in] conflict_version
+  ///   The conflict version of the negotiation
+  /// \param[in] sequence
+  ///   The sequence of participant ids. Follows convention of other sequences
+  ///   (ie. The last ParticipantId is the owner of the table)
+  ///
+  /// \return A TableView into what participants are proposing.
+  TableViewPtr table_view(
```
mxgrey, line 103: Since this function shouldn't modify anything about the internal state, we should be able to make this a `const` member function.

rmf_traffic_ros2/src/rmf_traffic_ros2/schedule/Negotiation.cpp
```diff
@@ -949,6 +993,56 @@ class Negotiation::Implementation
 
     return std::make_shared<Handle>(for_participant, negotiators);
   }
+
+  void set_retained_history_count(uint count)
+  {
+    retained_history_count = count;
+  }
+
+  TableViewPtr table_view(
```
mxgrey, line 1002: This Implementation member function will also need to be made `const` when `Negotiation::table_view(~)` is made `const`.

rmf_traffic_ros2/src/rmf_traffic_ros2/schedule/Negotiation.cpp
```diff
@@ -949,6 +993,56 @@ class Negotiation::Implementation
...
+    uint64_t conflict_version,
+    const std::vector<ParticipantId>& sequence)
+  {
+    const auto negotiate_it = negotiations.find(conflict_version);
+    rmf_traffic::schedule::Negotiation::TablePtr table;
+
+    if (negotiate_it == negotiations.end())
+    {
+      auto history_it = history.find(conflict_version);
```
mxgrey, line 1011:
```suggestion
      const auto history_it = history.find(conflict_version);
```

rmf_traffic_ros2/src/rmf_traffic_ros2/schedule/Negotiation.cpp
```diff
@@ -949,6 +993,56 @@ class Negotiation::Implementation
...
+  {
+    retained_history_count = count;
+  }
+
+  TableViewPtr table_view(
+    uint64_t conflict_version,
+    const std::vector<ParticipantId>& sequence)
+  {
+    const auto negotiate_it = negotiations.find(conflict_version);
+    rmf_traffic::schedule::Negotiation::TablePtr table;
```
mxgrey, line 1007:
```suggestion
    rmf_traffic::schedule::Negotiation::ConstTablePtr table;
```

rmf_traffic_ros2/src/rmf_traffic_ros2/schedule/Negotiation.cpp
```diff
@@ -949,6 +993,56 @@ class Negotiation::Implementation
...
+        RCLCPP_WARN(node.get_logger(), "Conflict version %llu does not exist."
+          "It may have been successful and wiped", conflict_version);
+        return nullptr;
+      }
+      
+      table = history_it->second.table(sequence);
+    }
+    else
+    {
+      auto& room = negotiate_it->second.room;
```
mxgrey, line 1023:
```suggestion
      const auto& room = negotiate_it->second.room;
```

rmf_traffic_ros2/src/rmf_traffic_ros2/schedule/Negotiation.cpp
```diff
@@ -949,6 +993,56 @@ class Negotiation::Implementation
...
+      Negotiation& negotiation = room.negotiation;
```
mxgrey, line 1024:
```suggestion
      const Negotiation& negotiation = room.negotiation;
```

mxgrey (changes requested):
This is looking really great!

There are just a few loose threads left to tie up.

rmf_traffic/src/rmf_traffic/schedule/Negotiation.cpp
ddengster: fixed a7c6ccb

rmf_traffic_ros2/include/rmf_traffic_ros2/schedule/Negotiation.hpp
ddengster: fixed a7c6ccb

rmf_traffic_ros2/src/rmf_traffic_ros2/schedule/Negotiation.cpp
ddengster: fixed a7c6ccb

rmf_traffic_ros2/src/rmf_traffic_ros2/schedule/Negotiation.cpp
ddengster: fixed a7c6ccb

rmf_traffic_ros2/src/rmf_traffic_ros2/schedule/Negotiation.cpp
ddengster: fixed a7c6ccb

rmf_traffic_ros2/src/rmf_traffic_ros2/schedule/Negotiation.cpp
ddengster: fixed a7c6ccb

rmf_traffic_ros2/src/rmf_traffic_ros2/schedule/Negotiation.cpp
ddengster: fixed a7c6ccb

mxgrey (approved):
Thanks for adding this feature!

PR #141, opened 2020-07-29
Add a warning about using Fast RTPS and Fast DDS

gbiggs:
RMF is known to have problems with Fast RTPS/Fast DDS on Foxy. This adds a warning if the user has not changed the default RMW implementation, or if they have manually set it to `rmw_fastrtps_cpp`.

This should be merged into the `foxy` branch, not `main`.

### DIFF
--- a/rmf_traffic_ros2/CMakeLists.txt
+++ b/rmf_traffic_ros2/CMakeLists.txt
@@ -19,6 +19,7 @@ find_package(rmf_traffic_msgs REQUIRED)
 find_package(rmf_fleet_msgs REQUIRED)
 find_package(Eigen3 REQUIRED)
 find_package(rclcpp REQUIRED)
+find_package(rcpputils REQUIRED)
 
 if(BUILD_TESTING)
   find_package(rmf_cmake_uncrustify REQUIRED)
--- a/rmf_traffic_ros2/CMakeLists.txt
+++ b/rmf_traffic_ros2/CMakeLists.txt
@@ -19,7 +19,6 @@ find_package(rmf_traffic_msgs REQUIRED)
 find_package(rmf_fleet_msgs REQUIRED)
 find_package(Eigen3 REQUIRED)
 find_package(rclcpp REQUIRED)
-find_package(rcpputils REQUIRED)
 
 if(BUILD_TESTING)
   find_package(rmf_cmake_uncrustify REQUIRED)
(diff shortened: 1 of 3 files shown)


rmf_traffic_ros2/CMakeLists.txt
```diff
@@ -19,6 +19,7 @@ find_package(rmf_traffic_msgs REQUIRED)
 find_package(rmf_fleet_msgs REQUIRED)
 find_package(Eigen3 REQUIRED)
 find_package(rclcpp REQUIRED)
+find_package(rcpputils REQUIRED)
```
mxgrey, line 22: Is there a particular motivation to bring in this dependency instead of using the [STL std::getenv](https://en.cppreference.com/w/cpp/utility/program/getenv) function? I suspect `rcpputils::get_env_var` is probably just wrapping that function in `std::string`, but it's easy enough for us to cast from `char*` to `std::string` without bringing in a new dependency.

Granted, this is probably a very easy dependency to satisfy since we're already depending on `rclcpp`, but I usually prefer to err on the side of fewer dependencies when it comes to C++.

rmf_traffic_ros2/src/rmf_traffic_schedule/ScheduleNode.cpp
```diff
@@ -293,6 +295,16 @@ ScheduleNode::ScheduleNode()
         }
       }
     });
+
+  // Warning for Foxy users to not use rmw_fastrtps_cpp
+  std::string rmw_env_var = rcpputils::get_env_var("RMW_IMPLEMENTATION");
+  if (rmw_env_var.empty() || rmw_env_var == "rmw_fastrtps_cpp")
+  {
+    RCLCPP_WARN(get_logger(), "RMF is known not to work correctly when using "
```
mxgrey, line 303:
```suggestion
    RCLCPP_WARN(get_logger(), "RMF is known to not work correctly when using "
```

Maybe it's a regional dialect difference, but "known not to" sounds a bit off to my ear.

rmf_traffic_ros2/CMakeLists.txt
gbiggs: I was mainly copying what the `rmw_implementation` package does to get the environment variable. I'm happy to change it.

gbiggs:
It occurred to me that a more robust approach would be to check the actually-loaded RMW implementation. I've committed a change that does that, instead of looking at the environment variable.

PR #142, opened 2020-07-29
Feature/remove ros2 dep

aaronchongth:
* removes ROS 2 dependencies from `rmf_utils` and `rmf_traffic`
* only when `BUILD_TESTING` is called will it try to find `ament_cmake_catch2` and `rmf_cmake_uncrustify`

This should allow us to distribute `rmf_utils` and `rmf_traffic` binaries to be used on non-ROS-2 systems

### DIFF
--- a/rmf_traffic/CMakeLists.txt
+++ b/rmf_traffic/CMakeLists.txt
@@ -1,6 +1,6 @@
 cmake_minimum_required(VERSION 3.5.0)
 
-project(rmf_traffic)
+project(rmf_traffic VERSION 1.0.2)
 
 set(CMAKE_EXPORT_COMPILE_COMMANDS on)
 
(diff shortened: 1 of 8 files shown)


aaronchongth:
@marcoag, was wondering what your thoughts about this were. We had a chat about this yesterday, and you mentioned that we would then not be able to use the ROS build farms, however we would need a non-ROS build farm for SOSS some time in the future.

A specific use case for the binaries would be to have it available for non-ROS2 robots, that are integrating with RMF. Let me know if you had more thoughts about this.

gbiggs:
What prevents us using the ROS build farm? If they are still colcon packages, then the build farm should be capable of building them even with pure CMake.

marcoag:
Should be fine on the ros buildfarm side, I might just have to add some patches on the release repo. Keeping it as a colcon package helps so the dependencies can be managed here through the `package.xml` file. The only thing to take into account is that the package will still be released under a certain ros release with the name`ros-<rosreleasename>-<packagename>`.

Issue #145, opened 2020-08-04
Fleet keeps going back-and-forth / deadlock during replanning and negotiation


Briancbn:
This might actually indirectly solve #144, but maybe a different problem in nature.

We notice that running [rmf_demos:office_loop_conflict.launch.xml](https://github.com/osrf/rmf_demos/blob/master/demos/launch/include/office/office_loop_conflict.launch.xml), there are two major problem when resolving conflict

- Conflict takes a pretty long time to resolve, negotiation has to happen several times (sometimes > 6).
- In the process fleets keep going back and forth undecisively. (it seems that after replanning, the fleet will typically reverse before continuing the task)
- Deadlock when head-to-head conflict happens on a single lane

Test was done in simulation. This happens in the world we set up as well. Is this an expected behaviour out of fleet adapter? or is there any way to avoid this problem. If this problem is known, what is the plan currently in addressing this.

mxgrey:
Whenever reporting a behavioral issue, please try to provide video recordings of what's happening. Verbal descriptions are not likely to be sufficient to identify the problem. It will also help if you can share what commands you're running to produce the issue.

Are the deadlocks happening around doors or lifts? If so, there are some known issues with negotiating through the use of such infrastructure, and we're already working on tying up those loose threads. Trying to negotiate through doorways is the most common cause of the back-and-forth behavior that you're describing.

I'm surprised to hear that head-to-head conflicts along a non-door lane would result in a deadlock, but possible cause I can think of is if one of the vehicles has a loop request goal waypoint in the middle of the deadlocked zone. The loop request is a half-baked task request that was only meant for demo purposes, so there are some rules for using it safely, like the endpoints should only be leaf nodes on the navigation graph.

mxgrey:
One other possible cause of a head-to-head deadlock is an inconsistency between the vicinity parameters of the fleet adapter and the parameters of the simulated robot. Unfortunately the current pipeline has two sources of truth for that. I'll double-check those parameters as soon as I get the chance.

Briancbn:
Thanks @mxgrey for the prompt reply

> Whenever reporting a behavioral issue, please try to provide video recordings of what's happening. Verbal descriptions are not likely to be sufficient to identify the problem.

- Keep going back and forth ([video](https://drive.google.com/file/d/1uUA6Z9UhHndDCpEIpRLMAcP41D2C18Sm/view?usp=sharing))
[...]

PR #149, opened 2020-08-13
Use SensorDataQos for state aggregator subscriber

Briancbn:
Signed-off-by: Chen Bainian <brian97cbn@gmail.com>
Related to #145, but doesn't fix all the problems.
We found out that sometimes one of the vehicles' `/fleet_states` doesn't get updated properly or missed during the operation. I think this was caused by the two messages are two close to each other in time, which somehow one of them gets missed by the state_aggregator subscriber and sometimes multiple times, causing a 2-4s delay! It turns out fine in `ros2 topic echo` for the `/robot_state`, but bad for `/fleet_states`. So I used the SensorDataQos instead, greatly improved the updates in `/fleet_states`. You can see the improvement in the following videos. We also tested under higher frequency for the fleets.

[Before Video](https://drive.google.com/file/d/1VidXTm12wjMR2JMWHxumoxfdu-x90EQC/view?usp=sharing), [After Video](https://drive.google.com/file/d/1B-qw4Lkti6TTqJElqnmNIzjXYS-8yjAx/view?usp=sharing)

It occurs occasionally for the office demo(very obvious under high publishing frequency), but very bad in our world.

I am not sure this is the best way, or whether state_aggregator will still exists in the new API or maybe a component implementation, please let me know.

### DIFF
--- a/rmf_fleet_adapter/src/robot_state_aggregator/main.cpp
+++ b/rmf_fleet_adapter/src/robot_state_aggregator/main.cpp
@@ -56,12 +56,13 @@ private:
   : rclcpp::Node("robot_state_aggregator")
   {
     const auto default_qos = rclcpp::SystemDefaultsQoS();
+    const auto sensor_qos = rclcpp::SensorDataQoS();
 
     _fleet_state_pub = create_publisher<FleetState>(
       rmf_fleet_adapter::FleetStateTopicName, default_qos);
 
     _robot_state_sub = create_subscription<RobotState>(
-      "robot_state", default_qos,
+      "robot_state", sensor_qos,
       [&](RobotState::UniquePtr msg)
       {
         _robot_state_update(std::move(msg));


rmf_fleet_adapter/src/robot_state_aggregator/main.cpp
```diff
@@ -55,13 +55,14 @@ class RobotStateAggregator : public rclcpp::Node
   RobotStateAggregator()
   : rclcpp::Node("robot_state_aggregator")
   {
-    const auto default_qos = rclcpp::SystemDefaultsQoS();
+    const auto pub_qos = rclcpp::SystemDefaultsQoS();
```
Yadunund, line 58:
```suggestion
    const auto default_qos = rclcpp::SystemDefaultsQoS();
```

rmf_fleet_adapter/src/robot_state_aggregator/main.cpp
```diff
@@ -55,13 +55,14 @@ class RobotStateAggregator : public rclcpp::Node
...
+    const auto sub_qos = rclcpp::SensorDataQoS();
 
     _fleet_state_pub = create_publisher<FleetState>(
-      rmf_fleet_adapter::FleetStateTopicName, default_qos);
+      rmf_fleet_adapter::FleetStateTopicName, pub_qos);
```
Yadunund, line 62:
```suggestion
      rmf_fleet_adapter::FleetStateTopicName, default_qos);
```

rmf_fleet_adapter/src/robot_state_aggregator/main.cpp
```diff
@@ -55,13 +55,14 @@ class RobotStateAggregator : public rclcpp::Node
   RobotStateAggregator()
   : rclcpp::Node("robot_state_aggregator")
   {
-    const auto default_qos = rclcpp::SystemDefaultsQoS();
+    const auto pub_qos = rclcpp::SystemDefaultsQoS();
+    const auto sub_qos = rclcpp::SensorDataQoS();
```
Yadunund, line 59:
```suggestion
    const auto sensor_qos = rclcpp::SensorDataQoS();
```

rmf_fleet_adapter/src/robot_state_aggregator/main.cpp
```diff
@@ -55,13 +55,14 @@ class RobotStateAggregator : public rclcpp::Node
...
 
     _fleet_state_pub = create_publisher<FleetState>(
-      rmf_fleet_adapter::FleetStateTopicName, default_qos);
+      rmf_fleet_adapter::FleetStateTopicName, pub_qos);
 
     _robot_state_sub = create_subscription<RobotState>(
-      "robot_state", default_qos,
+      "robot_state", sub_qos,
```
Yadunund, line 65:
```suggestion
      "robot_state", sensor_qos,
```

Yadunund (changes requested):
Hi @Briancbn ,

Thanks for identifying the problem with missed `RobotState` messages and for the suitable fix. I can confirm that switching to `SensorDataQoS` settings alleviates the problem. We can merge this in after addressing the nitpick comments on variable naming.

Briancbn:
Thanks. Updated in the latest commit and squashed.

Yadunund (approved):

PR #155, opened 2020-08-20
Tests for maximum_cost_estimate

gbiggs:
Adds tests relating to the `maximum_cost_estimate` option for planners.

Tests for setting and getting the option are provided.

Tests for the option's impact on planning are provided. These attempt to test rational allowable ranges. Currently the final test (`Cost limit just above actual required limit succeeds`) is failing.

The graph is probably overkill.

I'd welcome any suggests to make these tests more succinct, as well as any suggestions on what else should be tested.

Signed-off-by: Geoffrey Biggs <gbiggs@killbots.net>

### DIFF
--- a/rmf_traffic/test/unit/agv/test_Planner.cpp
+++ b/rmf_traffic/test/unit/agv/test_Planner.cpp
@@ -598,7 +598,7 @@ SCENARIO("Maximum Cost Estimates", "[maximum_cost_estimate]")
         options.maximum_cost_estimate(required_cost - 1e-6);
         planner.set_default_options(options);
         plan = planner.plan(start, goal);
-        CHECK(!plan);
+        CHECK_FALSE(plan);
       }
 
       THEN("Cost limit just above actual required limit succeeds")
--- a/rmf_traffic/test/unit/agv/test_Planner.cpp
+++ b/rmf_traffic/test/unit/agv/test_Planner.cpp
@@ -607,7 +607,7 @@ SCENARIO("Maximum Cost Estimates", "[maximum_cost_estimate]")
         options.maximum_cost_estimate(required_cost + 1e-6);
         planner.set_default_options(options);
         plan = planner.plan(start, goal);
-        CHECK(!plan);
+        CHECK(plan);
       }
     }
   }
--- a/rmf_traffic/test/unit/agv/test_Planner.cpp
+++ b/rmf_traffic/test/unit/agv/test_Planner.cpp
@@ -589,7 +589,7 @@ SCENARIO("Maximum Cost Estimates", "[maximum_cost_estimate]")
         options.maximum_cost_estimate(required_cost / 2);
         planner.set_default_options(options);
         plan = planner.plan(start, goal);
-        CHECK(!plan);
+        CHECK_FALSE(plan);
       }
 
       THEN("Cost limit just below actual required limit fails")
--- a/rmf_traffic/test/unit/agv/test_Planner.cpp
+++ b/rmf_traffic/test/unit/agv/test_Planner.cpp
@@ -458,6 +458,159 @@ SCENARIO("Test Options", "[options]")
     CHECK(*default_options.interrupt_flag());
   }
 
+  Planner::Options default_maxcostestimate(nullptr);
+  WHEN("Maximum cost estimate is default after default construction")
+  {
+    // Converting an optional to a bool returns false if it's empty, which is
+    // the default for constructing Options
+    CHECK(!default_maxcostestimate.maximum_cost_estimate());
+  }
+
+  double local_maximum_cost_estimate(3.14159);
+  Planner::Options set_maxcostestimate(
+    nullptr,
+    Planner::Options::DefaultMinHoldingTime,
+    interrupt_flag,
+    local_maximum_cost_estimate);
+  WHEN("Maximum cost estimate is set after construction")
+  {
+    CHECK(set_maxcostestimate.maximum_cost_estimate().value()
+      == Approx(local_maximum_cost_estimate).margin(1e-6));
+  }
+
+  WHEN("Set the maximum cost estimate")
+  {
+    local_maximum_cost_estimate = 42;
+    set_maxcostestimate.maximum_cost_estimate(local_maximum_cost_estimate);
+    CHECK(set_maxcostestimate.maximum_cost_estimate().value()
+      == Approx(local_maximum_cost_estimate).margin(1e-6));
+  }
+}
+
+SCENARIO("Maximum Cost Estimates", "[maximum_cost_estimate]")
+{
+  GIVEN("A graph")
+  {
+    const std::string test_map_name = "test_map";
+    rmf_traffic::agv::Graph graph;
+    graph.add_waypoint(test_map_name, {-5, -5}).set_passthrough_point(true); // 0
+    graph.add_waypoint(test_map_name, { 0, -5}).set_passthrough_point(true); // 1
+    graph.add_waypoint(test_map_name, { 5, -5}).set_passthrough_point(true); // 2
+    graph.add_waypoint(test_map_name, {10, -5}).set_passthrough_point(true); // 3
+    graph.add_waypoint(test_map_name, {-5, 0}); // 4
+    graph.add_waypoint(test_map_name, { 0, 0}); // 5
+    graph.add_waypoint(test_map_name, { 5, 0}); // 6
+    graph.add_waypoint(test_map_name, {10, 0}).set_passthrough_point(true); // 7
+    graph.add_waypoint(test_map_name, {10, 4}).set_passthrough_point(true); // 8
+    graph.add_waypoint(test_map_name, { 0, 8}).set_passthrough_point(true); // 9
+    graph.add_waypoint(test_map_name, { 5, 8}).set_passthrough_point(true); // 10
+    graph.add_waypoint(test_map_name, {10, 12}).set_passthrough_point(true); // 11
+    graph.add_waypoint(test_map_name, {12, 12}).set_passthrough_point(true); // 12
+    CHECK(graph.num_waypoints() == 13);
+
+    auto add_bidir_lane = [&](const std::size_t w0, const std::size_t w1)
+      {
+        graph.add_lane(w0, w1);
+        graph.add_lane(w1, w0);
+      };
+
+    add_bidir_lane(0, 1);
+    add_bidir_lane(1, 2);
+    add_bidir_lane(2, 3);
+    add_bidir_lane(1, 5);
+    add_bidir_lane(3, 7);
+    add_bidir_lane(4, 5);
+    add_bidir_lane(6, 10);
+    add_bidir_lane(7, 8);
+    add_bidir_lane(9, 10);
+    add_bidir_lane(10, 11);
+    add_bidir_lane(5, 9);
+    add_bidir_lane(11, 12);
+
+    const rmf_traffic::Profile profile = create_test_profile(UnitCircle);
+    const rmf_traffic::agv::VehicleTraits traits(
+      {0.7, 0.3}, {1.0, 0.45}, profile);
+    auto interrupt_flag = std::make_shared<bool>(false);
+
+    WHEN("Goal from 12->1")
+    {
+      const rmf_traffic::Time time = std::chrono::steady_clock::now();
+      const auto start = rmf_traffic::agv::Planner::Start{time, 12, 0.0};
+      const auto goal = rmf_traffic::agv::Planner::Goal{1};
+
+      THEN("No cost limit succeeds")
+      {
+        auto options = rmf_traffic::agv::Planner::Options{nullptr};
+        rmf_traffic::agv::Planner planner{
+          rmf_traffic::agv::Planner::Configuration{graph, traits},
+          options
+        };
+        auto plan = planner.plan(start, goal);
+        REQUIRE(plan);
+        REQUIRE(plan.cost_estimate());
+      }
+
+      THEN("Cost limit of zero fails")
+      {
+        auto options = rmf_traffic::agv::Planner::Options{
+          nullptr,
+          rmf_traffic::agv::Planner::Options::DefaultMinHoldingTime,
+          interrupt_flag,
+          0 // Maximum cost estimate must be 0
+        };
+        rmf_traffic::agv::Planner planner{
+          rmf_traffic::agv::Planner::Configuration{graph, traits},
+          options
+        };
+        auto plan = planner.plan(start, goal);
+        CHECK(!plan);
+      }
+    }
+
+    WHEN("Goal from 12->1 and known required cost limit")
+    {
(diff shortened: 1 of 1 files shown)


rmf_traffic/test/unit/agv/test_Planner.cpp
```diff
@@ -458,6 +458,159 @@ SCENARIO("Test Options", "[options]")
     CHECK(*default_options.interrupt_flag());
   }
 
+  Planner::Options default_maxcostestimate(nullptr);
+  WHEN("Maximum cost estimate is default after default construction")
+  {
+    // Converting an optional to a bool returns false if it's empty, which is
+    // the default for constructing Options
+    CHECK(!default_maxcostestimate.maximum_cost_estimate());
```
mxgrey, line 466:
```suggestion
    CHECK_FALSE(default_max_cost_estimate_options.maximum_cost_estimate());
```

Catch2 provides `CHECK_FALSE` which is the recommended way to test for false expectations. It makes the readout more clear when the test fails, and I think it makes the test expectations more explicit.

rmf_traffic/test/unit/agv/test_Planner.cpp
```diff
@@ -458,6 +458,159 @@ SCENARIO("Test Options", "[options]")
     CHECK(*default_options.interrupt_flag());
   }
 
+  Planner::Options default_maxcostestimate(nullptr);
```
mxgrey, line 461:
```suggestion
  Planner::Options default_max_cost_estimate_options(nullptr);
```

I'd like to discourage run-on words. I think that reduces readability :sweat_smile:

rmf_traffic/test/unit/agv/test_Planner.cpp
```diff
@@ -458,6 +458,159 @@ SCENARIO("Test Options", "[options]")
...
+  WHEN("Maximum cost estimate is default after default construction")
+  {
+    // Converting an optional to a bool returns false if it's empty, which is
+    // the default for constructing Options
+    CHECK(!default_maxcostestimate.maximum_cost_estimate());
+  }
+
+  double local_maximum_cost_estimate(3.14159);
```
mxgrey, line 469:
```suggestion
  double local_maximum_cost_estimate = 3.14159;
```

I'd recommend the assignment approach to initialization since I think that's usually more clear.

rmf_traffic/test/unit/agv/test_Planner.cpp
```diff
@@ -458,6 +458,159 @@ SCENARIO("Test Options", "[options]")
...
+  Planner::Options set_maxcostestimate(
```
mxgrey, line 470:
```suggestion
  Planner::Options set_max_cost_estimate_options(
```

rmf_traffic/test/unit/agv/test_Planner.cpp
```diff
@@ -458,6 +458,159 @@ SCENARIO("Test Options", "[options]")
...
+    nullptr,
+    Planner::Options::DefaultMinHoldingTime,
+    interrupt_flag,
+    local_maximum_cost_estimate);
+  WHEN("Maximum cost estimate is set after construction")
+  {
+    CHECK(set_maxcostestimate.maximum_cost_estimate().value()
```
mxgrey, line 477:
```suggestion
    CHECK(set_max_cost_estimate_options.maximum_cost_estimate().value()
```

rmf_traffic/test/unit/agv/test_Planner.cpp
```diff
@@ -458,6 +458,159 @@ SCENARIO("Test Options", "[options]")
...
+      == Approx(local_maximum_cost_estimate).margin(1e-6));
```
mxgrey, line 478:
```suggestion
      == Approx(local_maximum_cost_estimate));
```

The default margin should be okay here since we're expecting an exact binary match between these values.

rmf_traffic/test/unit/agv/test_Planner.cpp
```diff
@@ -458,6 +458,159 @@ SCENARIO("Test Options", "[options]")
...
+  }
+
+  WHEN("Set the maximum cost estimate")
+  {
+    local_maximum_cost_estimate = 42;
+    set_maxcostestimate.maximum_cost_estimate(local_maximum_cost_estimate);
+    CHECK(set_maxcostestimate.maximum_cost_estimate().value()
+      == Approx(local_maximum_cost_estimate).margin(1e-6));
```
mxgrey, line 486:
```suggestion
      == Approx(local_maximum_cost_estimate));
```

rmf_traffic/test/unit/agv/test_Planner.cpp
```diff
@@ -458,6 +458,159 @@ SCENARIO("Test Options", "[options]")
...
+  WHEN("Maximum cost estimate is set after construction")
+  {
+    CHECK(set_maxcostestimate.maximum_cost_estimate().value()
+      == Approx(local_maximum_cost_estimate).margin(1e-6));
+  }
+
+  WHEN("Set the maximum cost estimate")
+  {
+    local_maximum_cost_estimate = 42;
+    set_maxcostestimate.maximum_cost_estimate(local_maximum_cost_estimate);
```
mxgrey, line 484:
```suggestion
    set_max_cost_estimate_options.maximum_cost_estimate(local_maximum_cost_estimate);
```

rmf_traffic/test/unit/agv/test_Planner.cpp
```diff
@@ -458,6 +458,159 @@ SCENARIO("Test Options", "[options]")
...
+    CHECK(set_maxcostestimate.maximum_cost_estimate().value()
```
mxgrey, line 485:
```suggestion
    CHECK(set_max_cost_estimate_options.maximum_cost_estimate().value()
```

rmf_traffic/test/unit/agv/test_Planner.cpp
```diff
@@ -458,6 +458,159 @@ SCENARIO("Test Options", "[options]")
...
+
+      THEN("No cost limit succeeds")
+      {
+        auto options = rmf_traffic::agv::Planner::Options{nullptr};
+        rmf_traffic::agv::Planner planner{
+          rmf_traffic::agv::Planner::Configuration{graph, traits},
+          options
+        };
+        auto plan = planner.plan(start, goal);
+        REQUIRE(plan);
```
mxgrey, line 549: I'd suggest renaming `plan` to `result` since the result of attempting to produce a plan might or might not have actually succeeded in producing a plan.

Using other code in this test as a reference might make this matter confusing because the API went through a major revision since the time that most of this test file was written, but `Planner::plan(~)` produces a `Planner::Result` which is kind of like an `optional<Plan>` but with a lot more utilities, like being able to change some planning options and continue trying to plan.

rmf_traffic/test/unit/agv/test_Planner.cpp
```diff
@@ -458,6 +458,159 @@ SCENARIO("Test Options", "[options]")
...
+        REQUIRE(plan.cost_estimate());
```
mxgrey, line 550: There's no reason to expect the `Planner::Result` to still have a cost estimate after it has finished finding a plan. The `cost_estimate` field is only meant to be used when a planning effort has been interrupted; it's meant to help you decide whether or not you should continue bothering to search for a plan.

To get the final cost of a plan that's been successfully found, you could do `->get_cost()` on a `Planner::Result` where `->` provides access into a `Plan` object when a `Planner::Result` has successfully found a plan.

rmf_traffic/test/unit/agv/test_Planner.cpp
```diff
@@ -458,6 +458,159 @@ SCENARIO("Test Options", "[options]")
...
+          rmf_traffic::agv::Planner::Options::DefaultMinHoldingTime,
+          interrupt_flag,
+          0 // Maximum cost estimate must be 0
+        };
+        rmf_traffic::agv::Planner planner{
+          rmf_traffic::agv::Planner::Configuration{graph, traits},
+          options
+        };
+        auto plan = planner.plan(start, goal);
+        CHECK(!plan);
```
mxgrey, line 566:
```suggestion
        CHECK_FALSE(plan);
```

rmf_traffic/test/unit/agv/test_Planner.cpp
```diff
@@ -458,6 +458,159 @@ SCENARIO("Test Options", "[options]")
...
+      const auto goal = rmf_traffic::agv::Planner::Goal{1};
+
+      auto options = rmf_traffic::agv::Planner::Options{nullptr};
+      rmf_traffic::agv::Planner planner{
+        rmf_traffic::agv::Planner::Configuration{graph, traits},
+        options
+      };
+      auto plan = planner.plan(start, goal);
+      REQUIRE(plan);
+      REQUIRE(plan.cost_estimate());
```
mxgrey, line 583: As explained earlier, this shouldn't be a requirement or expectation.

rmf_traffic/test/unit/agv/test_Planner.cpp
```diff
@@ -458,6 +458,159 @@ SCENARIO("Test Options", "[options]")
...
+      auto required_cost = plan.cost_estimate().value();
```
mxgrey, line 584:
```suggestion
      auto required_cost = plan->get_cost();
```

rmf_traffic/test/unit/agv/test_Planner.cpp
```diff
@@ -458,6 +458,159 @@ SCENARIO("Test Options", "[options]")
...
+
+      THEN("Cost limit half actual required limit fails")
+      {
+
```
mxgrey, line 588: Unnecessary whitespace

rmf_traffic/test/unit/agv/test_Planner.cpp
```diff
@@ -458,6 +458,159 @@ SCENARIO("Test Options", "[options]")
...
+        options.maximum_cost_estimate(required_cost / 2);
+        planner.set_default_options(options);
+        plan = planner.plan(start, goal);
+        CHECK(!plan);
```
mxgrey, line 592:
```suggestion
        CHECK_FALSE(plan);
```

rmf_traffic/test/unit/agv/test_Planner.cpp
```diff
@@ -458,6 +458,159 @@ SCENARIO("Test Options", "[options]")
...
+      }
+
+      THEN("Cost limit just below actual required limit fails")
+      {
+
+        options.maximum_cost_estimate(required_cost - 1e-6);
+        planner.set_default_options(options);
+        plan = planner.plan(start, goal);
+        CHECK(!plan);
```
mxgrey, line 601:
```suggestion
        CHECK_FALSE(plan);
```

rmf_traffic/test/unit/agv/test_Planner.cpp
```diff
@@ -458,6 +458,159 @@ SCENARIO("Test Options", "[options]")
...
+      }
+
+      THEN("Cost limit just above actual required limit succeeds")
+      {
+
+        options.maximum_cost_estimate(required_cost + 1e-6);
+        planner.set_default_options(options);
+        plan = planner.plan(start, goal);
+        CHECK(!plan);
```
mxgrey, line 610:
```suggestion
        CHECK(plan);
```

Looks like this expectation got accidentally flipped.

rmf_traffic/test/unit/agv/test_Planner.cpp
gbiggs: Sorry, I missed changing that one! I usually don't use them myself (this is why we have auto-complete, right?), but probably got lazy here...

rmf_traffic/test/unit/agv/test_Planner.cpp
```diff
@@ -458,6 +458,154 @@ SCENARIO("Test Options", "[options]")
...
+  WHEN("Maximum cost estimate is set after construction")
+  {
+    CHECK(set_max_cost_estimate_options.maximum_cost_estimate().value()
+      == Approx(local_maximum_cost_estimate));
+  }
+
+  WHEN("Set the maximum cost estimate")
+  {
+    local_maximum_cost_estimate = 42;
+    set_max_cost_estimate_options.maximum_cost_estimate(local_maximum_cost_estimate);
```
mxgrey, line 484: Maybe add

```
REQUIRE(set_max_cost_estimate_options.maximum_cost_estimate());
```

here too, for the same reason as above.

rmf_traffic/test/unit/agv/test_Planner.cpp
```diff
@@ -458,6 +458,154 @@ SCENARIO("Test Options", "[options]")
...
+    add_bidir_lane(7, 8);
+    add_bidir_lane(9, 10);
+    add_bidir_lane(10, 11);
+    add_bidir_lane(5, 9);
+    add_bidir_lane(11, 12);
+
+    const rmf_traffic::Profile profile = create_test_profile(UnitCircle);
+    const rmf_traffic::agv::VehicleTraits traits(
+      {0.7, 0.3}, {1.0, 0.45}, profile);
+    auto interrupt_flag = std::make_shared<bool>(false);
```
mxgrey, line 533: Just a remark: When creating the `Planner::Options` below, you could just pass in a `nullptr` instead of actually creating an `interrupt_flag` that's always false. If a `nullptr` is given for the interrupt flag, then the planner will just run without being interrupted.

It doesn't really matter, though; it won't impact the test.

rmf_traffic/test/unit/agv/test_Planner.cpp
```diff
@@ -458,6 +458,154 @@ SCENARIO("Test Options", "[options]")
...
+
+    WHEN("Goal from 12->1")
+    {
+      const rmf_traffic::Time time = std::chrono::steady_clock::now();
+      const auto start = rmf_traffic::agv::Planner::Start{time, 12, 0.0};
+      const auto goal = rmf_traffic::agv::Planner::Goal{1};
+
+      THEN("No cost limit succeeds")
```
mxgrey, line 541:
```suggestion
      THEN("Success when there is no cost limit")
```

I kept reading "No cost limit succeeds" as saying that no cost limit would result in a success, so I think this rephrasing should make it a bit more clear.

rmf_traffic/test/unit/agv/test_Planner.cpp
```diff
@@ -458,6 +458,154 @@ SCENARIO("Test Options", "[options]")
...
+        options
+      };
+      auto result = planner.plan(start, goal);
+      REQUIRE(result);
+      auto required_cost = result->get_cost();
+
+      THEN("Cost limit half actual required limit fails")
+      {
+        options.maximum_cost_estimate(required_cost / 2);
+        planner.set_default_options(options);
```
mxgrey, line 587: Instead of changing the default options for the planner, you should be able to just call

```
result = planner.plan(start, goal, options);
```

mxgrey (approved):
Thanks for iterating on this!

I have a few more final remarks, but they're largely inconsequential :+1:

PR #156, opened 2020-08-24
Fix/estimate waypoint

Yadunund:
A minor tweak to use `compute_plan_starts()` inside `estimate_waypoint()` when `level_name` is published in the `RobotState` message.

### DIFF
--- a/rmf_fleet_adapter/src/full_control/estimation.cpp
+++ b/rmf_fleet_adapter/src/full_control/estimation.cpp
@@ -214,6 +214,12 @@ void estimate_state(
     TravelInfo& info)
 {
   std::string last_known_map = l.level_name;
+  if(!last_known_map.empty())
+  {
+    info.updater->update_position(last_known_map, {l.x, l.y, l.yaw});
+    return;
+  }
+
   if (info.last_known_wp)
   {
     const auto& last_known_wp = info.graph->get_waypoint(*info.last_known_wp);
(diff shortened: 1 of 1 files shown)


rmf_fleet_adapter/src/full_control/estimation.cpp
```diff
@@ -214,6 +214,12 @@ void estimate_state(
     TravelInfo& info)
 {
   std::string last_known_map = l.level_name;
+  if(!last_known_map.empty())
+  {
+    info.updater->update_position(last_known_map, {l.x, l.y, l.yaw});
```
mxgrey, line 219: I'm concerned that this will effectively erase information about the last known waypoint, and it's not clear to me if that's a good thing or a bad thing.

Are there specific problematic scenarios that this has managed to help with? If so, maybe the implementation of this entire function needs to be reconsidered.

rmf_fleet_adapter/src/full_control/estimation.cpp
```diff
@@ -260,6 +266,19 @@ void estimate_waypoint(
...
+  if(!last_known_map.empty())
+  {
+    const auto starts = rmf_traffic::agv::compute_plan_starts(
+      *info.graph,
+      last_known_map,
+      {l.x, l.y, l.yaw},
+      rmf_traffic_ros2::convert(node->now()));
+
+    assert(starts.size() > 0);
+    info.updater->update_position(starts[0].waypoint(), l.yaw);
```
mxgrey, line 278: In the absence of any information besides the location information, it's best to use [this `update_position` signature](https://github.com/osrf/rmf_core/blob/4180a944e7497a3dd6b08f17280706ce6aca8a30/rmf_fleet_adapter/include/rmf_fleet_adapter/agv/RobotUpdateHandle.hpp#L79) because it will be effectively the same as what you're doing here, except it will take all possible starts into account.

That being said, this also throws away information about the last known map and last known waypoint. If this is an improvement on the previous implementation, then we should probably be completely throwing out the old implementation.

mxgrey (commented):
If you can elaborate on the rationale behind these changes and provide examples of situations where the these changes improve the behavior, that would help a lot in evaluating this PR.

rmf_fleet_adapter/src/full_control/estimation.cpp
Yadunund: Hi @mxgrey, I figured the new changes here and below would improve estimations for situations where the robot transitions from a waypoint on one level to another. I made these changes initially while debugging the issue with robots ignoring their initial lane conditions at  times but that problem was solved by a different fix. If these changes are not useful, I can close this PR.

mxgrey: I don't want to jump to any conclusions about whether they're useful or not.

But if these changes  **are** substantive improvements, then I think that suggests many of the assumptions that I made in this implementation were bad, and the whole implementation should be reconsidered.

I'm very open to that possibility; I just think it would be good to have some experimental validation of that before diving into that rabbit hole.

Yadunund: That's a fair point @mxgrey. I'll work on experimental validation of this before pursuing any further.

Yadunund:
Closing issue for now.

PR #159, opened 2020-08-24
Improve heuristic to account for events

mxgrey:
This PR improves the planning heuristic to account for the time taken up by waiting on events, such as waiting on doors. Without taking that into account, the planner may try too hard to avoid waiting for doors, even if there isn't any better option.

The heuristic is still in need of some critical improvements, but this should at least be a step in the right direction.


Kevinskwk (approved):
It works great on my side, thanks!

PR #158, opened 2020-08-24
Add a mandatory waiting event

mxgrey:
The mandatory wait event is a generic placeholder for "events" that require the fleet adapter to simply wait for a period of time. Unlike other event types, this event does not expect the fleet adapter to do anything or wait for anything besides the elapsed amount of time.

PR #161, opened 2020-08-24
Keep better track of docking waypoint

mxgrey:
This PR should help robots to undock by keeping track of the waypoint that they most recently docked into.


mxgrey:
To simplify things, we should review and merge #158 first.

Yadunund (approved):
Robots no longer re-dock when exiting a docking station!

PR #163, opened 2020-08-24
Make planner better about changing floors

mxgrey:
Until now, the planner has not been very good about changing floors, with a high risk of various assertion failures when that happens.

This PR should resolve that issue for the simple case of vertically moving between floors using a lift. This should be made more robust in the future, which will be important for other cases of floor-changing, like ramps.


Kevinskwk (approved):
Working great, conflict checks and negotiations are taking place normally. Thanks!

PR #165, opened 2020-08-25
Feature/battery planning

Yadunund:
This PR introduces a new package, `rmf_battery` to model the battery drain for mobile robot participants in RMF.
In particular it introduces:
* `rmf_battery::agv::BatterySystem` class which is a container for battery properties of the vehicle.
* `rmf_battery::agv::MechanicalSystem` class which is a container for mechanical properties of the vehicle.
* `rmf_battery::agv::PowerSystem` class which is a container for power system/device properties of the vehicle.
* `rmf_battery::MotionPowerSink` pure abstract class which contains `compute_change_in_charge(rmf_traffic::Trajectory trajectory)`. A sample implementation is provided by `rmf_battery::agv::SimpleMotionPowerSink`
* `rmf_battery::DevicePowerSink` pure abstract class which contains `compute_change_in_charge(double run_time)`. A sample implementation is provided by `rmf_battery::agv::SimpleDevicePowerSink`

The expectation is for the user to use a combination of `MotionPowerSink` and `DevicePowerSink(s)` to compute the expected battery drain for various applications


rmf_battery/include/rmf_battery/EstimateBattery.hpp
```diff
@@ -0,0 +1,65 @@
...
+  /// \param[in] power_map
+  ///   An optional unordered map with keys representing names of power systems
+  ///   and values of trajectories during which the power system is active.
+  ///
+  /// \return The remaining charge at the end of the trajectory as a fraction of
+  ///   total battery capacity
+  virtual double compute_state_of_charge(
+    const rmf_traffic::Trajectory& trajectory,
+    const double initial_soc,
+    rmf_utils::optional<PowerMap> power_map = rmf_utils::nullopt) const = 0;
```
mxgrey, line 58: I don't understand why the value of this map would be a `rmf_traffic::Trajectory`. If the function is supposed to compute the state of charge based on a given trajectory, what does the function do with these additional trajectory values?

Based on what I'm seeing in `SimpleBatteryEstimator.cpp`, it looks like the only aspect of `rmf_traffic::Trajectory` that is being used by the `compute_state_of_charge` calculation is the ability to find time segments. I think this is an abuse of the `Trajectory` class, and the intention of the function becomes very confusing because the user is providing both a `trajectory` argument and an optional map of trajectories.

I don't know much about the math involved in state of charge, but I take it the total power drawn isn't just the [superposition](https://en.wikipedia.org/wiki/Superposition_principle) of the power drawn by moving plus the power drawn by the device? They need to be integrated simultaneously to get an accurate prediction?

In that case, I might suggest making a pure abstract interface for a `PowerSystem` with a function that takes in a `rmf_traffic::Time` and a `double initial_soc` value and returns a finite differential to be integrated. (The current `PowerSystem` class would need to be renamed, perhaps to `SimplePowerSystem`.)

Then we can provide a `SimpleLocomotion` power system that accounts for the power drawn by following a trajectory.  Users can add their own custom power systems for various tools or devices on their robot.

Then the `EstimateBattery` class wouldn't be needed anymore. It could be replaced by a function that accepts a collection of `std::unique_ptr<PowerSystem>` objects and integrates them over an interval of time.

mxgrey (commented):
I haven't gotten through everything yet, but I do have some feedback on the `EstimateBattery` class, which I think needs some reworking.

rmf_battery/include/rmf_battery/EstimateBattery.hpp
Yadunund: `rmf_traffic::Trajectory` is used to infer the time interval over which the `PowerSystem` is active. The original idea is that a robot can have multiple power systems (sensors, cleaning systems, etc) some of which are active throughout (sensors) while others may only be active for a subset of the trajectory. To cater to this, the map accepts a `Trajectory` for each `PowerSystem`. But you're right in that this may be a misguided use of the `Trajectory` class. Does it make sense to set the value of the map to be an `std::pair<rmf_traffic::Time, rmf_traffic::Time> instead? Here the first and second elements represent the start and end times of the power systems.

With regards to the way the state of charge is computed, the power drawn is actually superimposible. While the power drawn by locomotion needs to be computed over an integral, the power drawn by PowerSystems is simply the `wattage * duration active`. The sum of the powers drawn by these systems can simply be added to that drawn by the robot when following a trajectory. Do you still recommend making `PowerSystem` a pure abstract interface?

mxgrey: > With regards to the way the state of charge is computed, the power drawn is actually superimposible. 

Thanks for clarifying this. Just to make sure I understand correctly, is the ability to superimpose always legitimately applicable, or is it just an acceptable estimate for our use-case?

To make my concern clear: Should we consider the possibility that some users will have power systems that need to be simultaneously integrated in order to get a sufficiently accurate prediction? Or would that be an unreasonable demand from our users?

> Do you still recommend making PowerSystem a pure abstract interface?

This is much less of a concern if the power consumption is superimposible, but I wouldn't rule it out yet.

> Does it make sense to set the value of the map to be an `std::pair<rmf_traffic::Time, rmf_traffic::Time> instead? 

I gather that there isn't an actual need to calculate all the power consumption simultaneously. In that case, there's no reason to have a single function that tries to crunch through all the different power systems simultaneously, especially if the implementation needs to interpret a string to identify what power system model to use. I think it makes more sense to decouple all of this, one way or another.

If all power systems can truly be superimposed, here's what my suggestion would be for a new iteration (feel free to exercise creativity with the naming; I'm just using the first names that come to mind):

```
class LocomotionPowerSink
{
public:
  double compute_change_in_charge(const rmf_traffic::Trajectory& trajectory) const = 0;
};

class DevicePowerSink
{
public:
  double compute_change_in_charge(double run_time) const = 0;
};
```

We can then use these two abstract interfaces to communicate everything that a cleaning task factory would need to know for calculating its power consumption. The cleaning task factory signature might look something like this:

```
std::optional<CleaningTask> make_cleaning_task(
  const Planner::StartSet& starts,
  double initial_soc,
  std::size_t cleaning_start_wp,
  const rmf_traffic::Trajectory& cleaning_trajectory,
  std::size_t cleaning_end_wp,
  std::vector<std::size_t> charging_stations,
  const LocomotionPowerSink& locomotion,
  const DevicePowerSink& cleaning_tool, // power that gets consumed while cleaning
  const DevicePowerSink& ambient_devices // power that is always getting consumed
);
```

That function wouldn't need to be wrapped in an object the way it is in the current `CleaningTaskPlanner`. The `LocomotionPowerSink` would mostly use the implementation of the current `SimpleBatteryEstimator` class, but with the power map removed. The `DevicePowerSink` can use the existing `PowerSystem` + `BatterySystem` in its implementation, although we may want to make `PowerSystem` and `BatterySystem` into first-class citizens instead of nesting them inside of `SystemTraits`.

I think the big picture is that the `EstimateBattery` interface was effectively designed to have `SystemTraits` be used in all of its implementations (not explicitly, but implicitly by using string-identified power systems in its API), but that kind of coupling defeats the benefit of abstract interfaces. So instead of trying to group all the power systems together using the `SystemTraits` class, I propose that we keep the systems separate and leave it to the user to define their power systems and keep that all organized. Of course they're welcome to use our out-of-the-box implementations of these interfaces, but we shouldn't impose them.

Yadunund: Thanks for this extremely valuable feedback!

>Just to make sure I understand correctly, is the ability to superimpose always legitimately applicable, or is it just an acceptable estimate for our use-case?
Based on exhaustive literature review it appears to be acceptable to superimpose sources of power drain for this use case.

> I think it makes more sense to decouple all of this, one way or another.
I see the benefit of this approach. Thanks for the suggestion and examples of how to go about achieving the same.

> The cleaning task factory signature might look something like this
Here, is `CleaningTask` a class defined in a new namespace say `rmf_battery::tasks::CleaningTask`? If so, should it implement a `plan()` method to compute the cleaning plan or does the `make_cleaning_task()` method directly compute the plan which can then be accessed through a member function / drill operator?

The rest of your comments are crystal clear. I'll begin the refactor asap. Thanks again!

mxgrey: > Here, is CleaningTask a class defined in a new namespace say rmf_battery::tasks::CleaningTask?

I think it should go in a downstream package, not in `rmf_battery` at all. We should have `rmf_battery` focus on battery interfaces. We could put `CleaningTask` into an `rmf_task` package that depends on `rmf_battery`.

Yadunund:
Closing in favor of https://github.com/osrf/rmf_core/pull/217

PR #173, opened 2020-09-21
Update ROS2 workflow version from eloquent to foxy

mxgrey:


.github/workflows/build.yaml
```diff
@@ -2,9 +2,9 @@ name: build
...
 jobs:
   build:
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-20.04
     container:
-      image: docker://ros:eloquent-ros-base-bionic
+      image: docker://ros:foxy-ros-base-focal
       
     steps:
     - name: deps
```
luca-della-vedova, line 10: I guess this step can be removed as well?

mxgrey: Are you suggesting getting rid of the rosdep install? I'm not sure if `libfcl` and `libccd` are included in the base image of foxy :thinking: 

But even if they are, I think the pipeline is more robust to future dependency changes if we just keep this in.

luca-della-vedova: I meant the normal "deps" step that only installs g++8, but it probably doesn't matter that much

mxgrey: Oh sorry I got confused because GitHub had collapsed that part of the file, so it looked like you were referring to `rosdep install` :facepalm:

PR #181, opened 2020-10-02
Fix debug build error due to incorrect call to make_segment_num_error

mrushyendra:
Ensures correct arguments are passed to `make_segment_num_error`, so that debug builds work.


mxgrey (approved):
Whoops, thanks for catching this. I'm not sure how it's gone undetected this whole time :thinking:

PR #185, opened 2020-10-07
Fix/finish duration

aaronchongth:
* Removed all time related default arguments, especially those using `std::chrono::steady_clock::now()`
* Changed to using `finish_duration` instead of `finish_time`, so that everything is handled relatively to the planning start time
* Added arguments for relative start time to all relevant methods and objects
* Removed default constructor for `agv::State`
* Modified tests to work as before
* Throws exception when invalid battery state of charge values are passed.
* Some documentation


aaronchongth:
Closing in favour of https://github.com/osrf/rmf_core/pull/188

PR #188, opened 2020-10-12
Using rmf_traffic::agv::Plan::Start for holding waypoint and time information

aaronchongth:
Signed-off-by: Aaron Chong <aaronchongth@gmail.com>

* Removed all time related default arguments, especially those using std::chrono::steady_clock::now()
* Removed default constructor for agv::State
* Using `rmf_traffic::agv::Plan::Start` in `rmf_tasks::agv::State` to hold information for waypoint index and finish time (when the robot is available to accept the next task)
* Throws exception when invalid battery state of charge values are passed
* Added tests to check for exceptions for battery soc and threshold values in `State` and `StateConfig`
* Modified tests to work as before, added arguments for relative start time to all relevant methods and objects
* Some documentation

### DIFF
--- a/rmf_tasks/include/rmf_tasks/agv/State.hpp
+++ b/rmf_tasks/include/rmf_tasks/agv/State.hpp
@@ -24,44 +24,66 @@
 #include <rmf_utils/optional.hpp>
 
 #include <rmf_traffic/Time.hpp>
+#include <rmf_traffic/agv/Planner.hpp>
 
 namespace rmf_tasks {
 namespace agv {
 
+/// This state representation is used for task planning.
 class State
 {
 public:
 
   /// Constructor
   ///
-  /// \param[in] waypoint
+  /// \param[in] plan_start
+  ///   Current state's next start for new plans, includes the time which
+  ///   the plan can feasibly start, according to the finishing time of any
+  ///   tasks that the robot is currently performing.
+  ///
   /// \param[in] charging_waypoint
-  /// \param[in] finish_time
+  ///   The charging waypoint index of this robot.
+  ///
   /// \param[in] battery_soc
-  /// \param[in] threshold_soc
+  ///   Current battery state of charge of the robot. This value needs to be
+  ///   between 0.0 to 1.0.
   State(
-    std::size_t waypoint, 
+    rmf_traffic::agv::Plan::Start plan_start,
     std::size_t charging_waypoint,
-    rmf_traffic::Time finish_time = std::chrono::steady_clock::now(),
-    double battery_soc = 1.0);
-
-  State();
+    double battery_soc);
   
-  std::size_t waypoint() const;
+  /// The next plan start based on the current state.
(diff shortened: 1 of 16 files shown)


rmf_tasks/include/rmf_tasks/agv/State.hpp
```diff
@@ -24,45 +24,67 @@
...
-  /// \param[in] finish_time
+  ///   The charging waypoint index of this robot.
+  ///
   /// \param[in] battery_soc
-  /// \param[in] threshold_soc
+  ///   Current battery state of charge of the robot. This value needs to be
+  ///   between 0.0 to 1.0.
   State(
-    std::size_t waypoint, 
+    rmf_traffic::agv::Plan::Start plan_start,
```
Yadunund, line 51: I think the name `plan_start` is misleading and should be re-named to `location` as it is meant to reflect the expected finish location of the robot. This would align with the usage in `RobotContext` and `TaskManager` https://github.com/osrf/rmf_core/blob/89e758ac9b4fddc0302dac2d10d062ad7fadb477/rmf_fleet_adapter/src/rmf_fleet_adapter/TaskManager.hpp#L61

rmf_tasks/include/rmf_tasks/agv/TaskPlanner.hpp
```diff
@@ -127,6 +127,7 @@ class TaskPlanner
   /// Get the greedy planner based assignments for a set of initial states and 
   /// requests
   Assignments greedy_plan(
+    rmf_traffic::Time relative_start_time,
```
Yadunund, line 130: Would it be better to rename `relative_start_time` to `time_now` or `now`?

rmf_tasks/src/agv/TaskPlanner.cpp
```diff
@@ -614,6 +614,7 @@ class TaskPlanner::Implementation
   }
   
   Assignments complete_solve(
+    rmf_traffic::Time relative_start_time,
```
Yadunund, line 617: I might be missing something here but it looks like this new variable is not being used anywhere?

Yadunund (changes requested):
Thanks for switching the `State` container to use `rmf_traffic::agv::Planner::Start`. I have left a couple of comments for your consideration. It looks like the new `relative_start_time` field in `complete_solve` is not being used anywhere?

rmf_tasks/include/rmf_tasks/agv/State.hpp
aaronchongth: Hmm, I was using `location` previously but found the name and the type not complementing each other, as it is not immediately clear what `location` has to do with `Plan::Start`.

aaronchongth: For now I will prioritize how semantic the names are, we can decide down the road if it really becomes confusing :man_shrugging:

Yadunund (approved):
Thanks for making these changes so quickly! The added documentation is great too.

PR #190, opened 2020-10-14
Update schedule with docking itinerary

Yadunund:
Interim solution to publish the docking path of the robot to RMF schedule


aaronchongth (approved):
Thanks for wrangling with the itinerary! LGTM

PR #191, opened 2020-10-14
Use optional of reference wrapper

aaronchongth:
Signed-off-by: Aaron Chong <aaronchongth@gmail.com>

### DIFF
--- a/rmf_fleet_adapter/include/rmf_fleet_adapter/agv/RobotUpdateHandle.hpp
+++ b/rmf_fleet_adapter/include/rmf_fleet_adapter/agv/RobotUpdateHandle.hpp
@@ -28,6 +28,7 @@
 
 #include <vector>
 #include <memory>
+#include <functional>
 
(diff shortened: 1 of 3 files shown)


Yadunund (approved):
Thanks for addressing the UB. Pretty cool solution!

mxgrey:
A more classic alternative would be to return a `Participant*` which has a `nullptr` value if the object isn't available. That said, there's nothing wrong with this approach; the only downside is the `std::optional` uses up one extra byte to track whether the object is available.

aaronchongth:
I see, yeah that would be a good idea. I was trying to prevent any use of raw pointers in the API, thanks for the tip!

mxgrey:
It's okay to use raw pointers in the API as long the raw pointer is telling the user "You are allowed to have access to view/mutate this object, but you do not own its memory allocation". You should absolutely not use raw pointers if the user is expected to own and manage the memory allocation of the object.

PR #194, opened 2020-10-22
Port to fcl 0.6.1

ddengster:
As per title.

At the moment Foxy only has fcl-0.5, so this requires some adjustments to the build process; you'll have to pull in this [branch](https://github.com/ddengster/fcl/tree/0.6.1-fix-conservative-adv) into your workspace and colcon build it.

Relevant PR with fix: https://github.com/flexible-collision-library/fcl/pull/505


rmf_traffic/CMakeLists.txt
```diff
@@ -20,10 +20,7 @@ endif()
 
 include(GNUInstallDirs)
 
-include(FindPkgConfig)
-pkg_check_modules(PC_FCL REQUIRED fcl)
-pkg_check_modules(PC_CCD REQUIRED ccd)
-
+find_package(fcl 0.6 REQUIRED)
```
mxgrey, line 23:
```suggestion
find_package(fcl 0.6)
if(fcl_FOUND)
  set(FCL_LIBRARIES fcl)
  set(using_new_fcl true)
else()
  include(FindPkgConfig)
  pkg_check_modules(PC_FCL REQUIRED fcl)
  pkg_check_modules(PC_CCD REQUIRED ccd)
  set(FCL_LIBRARIES ${PC_FCL_LIBRARIES} ${PC_CCD_LIBRARIES})
  set(using_new_fcl)
endif()
```

rmf_traffic/CMakeLists.txt
```diff
@@ -44,7 +41,7 @@ if(BUILD_TESTING AND ament_cmake_catch2_FOUND AND rmf_cmake_uncrustify_FOUND)
     TIMEOUT 300)
   target_link_libraries(test_rmf_traffic
       rmf_traffic
-      ${PC_FCL_LIBRARIES}
+      fcl
       Threads::Threads
   )
```
mxgrey, line 47:
```suggestion
  if(using_new_fcl)
    target_compile_definitions(test_rmf_traffic PRIVATE RMF_TRAFFIC__USING_FCL_0_6)
  endif()
```

rmf_traffic/CMakeLists.txt
```diff
@@ -66,8 +63,7 @@ target_link_libraries(rmf_traffic
   PUBLIC
     rmf_utils::rmf_utils
   PRIVATE
-    ${PC_FCL_LIBRARIES}
-    ${PC_CCD_LIBRARIES}
+    fcl
 )
```
mxgrey, line 68:
```suggestion
if(using_new_fcl)
  target_compile_definitions(rmf_traffic PRIVATE RMF_TRAFFIC__USING_FCL_0_6)
endif()
```

rmf_traffic/package.xml
```diff
@@ -11,8 +11,7 @@
   <build_depend>rmf_utils</build_depend>
   <build_export_depend>rmf_utils</build_export_depend>
 
-  <build_depend>libccd-dev</build_depend>
-  <build_depend>libfcl-dev</build_depend>
+  <build_depend>fcl</build_depend>
```
mxgrey, line 14: I'm not sure what we can/should do about this to make it backwards compatible. One option might be to keep the old two `build_depend` tags in here for now. I don't think there's much harm in having redundant dependency declarations.

rmf_traffic/src/rmf_traffic/DetectConflict.cpp
```diff
@@ -23,9 +23,9 @@
 
 #include "DetectConflictInternal.hpp"
 
-#include <fcl/continuous_collision.h>
-#include <fcl/ccd/motion.h>
-#include <fcl/collision.h>
+#include <fcl/narrowphase/continuous_collision.h>
```
mxgrey, line 26: We can change these includes like:

```
#ifdef RMF_TRAFFIC__USING_FCL_0_6
#include <fcl/narrowphase/continuous_collision.h>
#include <fcl/math/motion/spline_motion.h>
#include <fcl/narrowphase/collision.h>

using SplineMotion = fcl::SplineMotion<double>;
using ContinuousCollisionRequest = fcl::ContinuousCollisionRequestd;
... etc ...

#else
#include <fcl/continuous_collision.h>
#include <fcl/ccd/motion.h>
#include <fcl/collision.h>

using SplineMotion = fcl::SplineMotion;
using ContinuousCollisionRequest = fcl::ContinuousCollisionRequest;
... etc ...

#endif // RMF_TRAFFIC__USING_FCL_0_6
```

We may also want to move some of the conversion functions whose APIs changed into these `#ifdef` blocks.

We'll also need to use this `#ifdef` pattern in all the other source files that use FCL. Alternatively, we can make a "[backward compatibility](https://github.com/dartsim/dart/blob/master/dart/collision/fcl/BackwardCompatibility.hpp)" header like what's done in DART and include that in all the different source files. If we take this approach, make sure to keep it as an internal header inside `src/rmf_traffic` rather than putting it in the public API in `include/rmf_traffic`.

mxgrey (changes requested):
Since this requires a non-trivial dependency change in the version of FCL, I'm concerned these changes could get stuck in PR limbo until eventually merge conflicts start to develop. To avoid that, I'd like to suggest that we tweak this PR to make it compatible with both 0.5 and 0.6 when compiling from source.

I've left inline comments to help guide the cross-compatibility effort.

rmf_traffic/src/rmf_traffic/DetectConflict.cpp
ddengster: 0c965de fcl 0.5 backwards compatibility

rmf_traffic/package.xml
ddengster: ebe339e

ddengster:
the upsream FCL PR https://github.com/flexible-collision-library/fcl/pull/505 has been merged

mxgrey (changes requested):
There are just two small changes needed in the cmake to ensure future-proofing and cross-compatibility.

mxgrey (approved):

PR #197, opened 2020-10-27
Fix lifetime of variant_duration variable

mrushyendra:
Minor fix to prevent redeclaration of `variant_duration` inside the if-block scope. Opening a separate PR for this so that it doesn't get lost with other changes to task planning, or in case the other changes do not get merged.


Yadunund (approved):
Thanks for identifying this fix!

PR #200, opened 2020-11-03
Added new Clean task, changed TaskProfile and SubmitTask to use explicit message definitions

aaronchongth:
* Changed `TaskType` value to use `uint32`, and changed naming scheme for uniformity
* Note, missing Patrol and Charge tasks


Yadunund (approved):
Thanks for updating these messages!

PR #201, opened 2020-11-04
add eval enum to submit task srv

tanyouliang95:
This will be parsed from user facing api srv call. default is `0: lowest-diff-cost `


Yadunund (approved):
LGTM!

PR #203, opened 2020-11-05
Change id field of Request to string and modify TaskPlanner to use internal ids instead

mrushyendra:
Modifies `request` to use string based id instead of size_t. `Node` generates its own custom ids for any requests that are added to it, for use by the `TaskPlanner`, instead of relying on the `request`'s `id` field.

### DIFF
--- a/rmf_task/src/rmf_task/agv/TaskPlanner.cpp
+++ b/rmf_task/src/rmf_task/agv/TaskPlanner.cpp
@@ -117,6 +117,7 @@ class TaskPlanner::Assignment::Implementation
 {
 public:
 
+  size_t request_id;
   rmf_task::RequestPtr request;
   State state;
   rmf_traffic::Time deployment_time;
@@ -124,17 +125,25 @@ public:
 
 //==============================================================================
 TaskPlanner::Assignment::Assignment(
+  size_t request_id,
   rmf_task::RequestPtr request,
   State state,
   rmf_traffic::Time deployment_time)
 : _pimpl(rmf_utils::make_impl<Implementation>(
       Implementation{
+        request_id,
         std::move(request),
         std::move(state),
         deployment_time
       }))
 {
   // Do nothing
+}
+
+//==============================================================================
+std::size_t TaskPlanner::Assignment::request_id() const
+{
+  return _pimpl->request_id;
 }
 
 //==============================================================================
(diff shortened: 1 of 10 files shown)


Yadunund:
@mrushyendra  can we resolve the new conflicts on this branch?

rmf_task/src/rmf_task/agv/TaskPlanner.cpp
```diff
@@ -370,6 +376,13 @@ struct Node
   double cost_estimate;
   rmf_traffic::Time latest_time;
   InvariantSet unassigned_invariants;
+  std::size_t next_available_internal_id = 1;
+
+  // ID 0 is reserved for charging tasks
+  std::size_t get_available_internal_id(bool charging_task = false)
+  {
+    return charging_task ? 0 : next_available_internal_id++;
```
Yadunund, line 384: If we're generating new ids for all requests why not generate them for charging tasks too? That way each charging task has a unique id?

rmf_task/src/rmf_task/agv/TaskPlanner.cpp
```diff
@@ -631,8 +646,19 @@ bool Filter::ignore(const Node& node)
 const rmf_traffic::Duration segmentation_threshold =
     rmf_traffic::time::from_seconds(1.0);
 
-} // anonymous namespace
+inline double compute_g_assignment(const TaskPlanner::Assignment& assignment)
```
Yadunund, line 649: Is there a reason for separating this block of code from `compute_g()`?

rmf_task/src/rmf_task/agv/TaskPlanner.cpp
mrushyendra: It was so that the `Filter` does not differentiate between 2 assignments that have different IDs for a charging task, but are otherwise the same.

rmf_task/src/rmf_task/agv/TaskPlanner.cpp
mrushyendra: It was to enable `compute_cost` and `compute_g` to use the same common function. If not, there is some duplication because `compute_cost` takes in an `Assignment` while `compute_g` takes in a `Node` that contains `AssignmentWrapper`s, so there becomes 2 separate functions to keep in sync with each other.

rmf_task/src/rmf_task/agv/TaskPlanner.cpp
Yadunund: Ah that makes sense. Thanks for the clarification.

rmf_task/src/rmf_task/agv/TaskPlanner.cpp
Yadunund: I was thinking `compute_g(Node)` can create `Assignments` from the `Node` and then call `compute_cost(Assignments)` on it. But this approach is valid too.

Yadunund (approved):
Thanks for making this switch and addressing all the comments!

PR #207, opened 2020-11-06
Reduce heuristic overestimation and ignore charging task costs in TaskPlanner

mrushyendra:
`g(n)` computes the cost of each task based on its `earliest_start_time`, whereas `h(n)`earlier did not. This meant that costs were overestimated, making A* less likely to pick the optimal solution. This PR modifies `h(n)` to take this into account. The heuristic should now underestimate the costs when the earliest start times for each task are similar/same (enforced by the segmentation_threshold), and should still perform well even if they are not, in all but some rare cases.

In addition, it modifies `g(n)` to ignore the cost of any charging tasks inserted into the plan, since the cost of charging is indirectly included in the tasks that follow. Right now it adds an additional field to `Assignment`. I'm not sure if there is a better way - any suggestions are welcome. (Edit: Actually I could look at the request field and specifically ignore it if it is a charging request using `dynamic_cast`)

Based on some [testing](https://github.com/mrushyendra/rmf_core/blob/task_planner_new/rmf_task/test/unit/agv/test_TaskPlanner.cpp) (not included in this PR), costs with both `greedy_solve` and `optimal_solve` decreased by about 2-30% in almost all cases, even after controlling for the decrease in costs from not considering charging tasks.

### DIFF
--- a/rmf_task/src/rmf_task/agv/TaskPlanner.cpp
+++ b/rmf_task/src/rmf_task/agv/TaskPlanner.cpp
@@ -120,21 +120,18 @@ public:
   rmf_task::RequestPtr request;
   State state;
   rmf_traffic::Time deployment_time;
-  bool include_in_cost;
 };
 
 //==============================================================================
 TaskPlanner::Assignment::Assignment(
   rmf_task::RequestPtr request,
   State state,
-  rmf_traffic::Time deployment_time,
-  bool include_in_cost)
+  rmf_traffic::Time deployment_time)
 : _pimpl(rmf_utils::make_impl<Implementation>(
       Implementation{
         std::move(request),
         std::move(state),
-        deployment_time,
-        include_in_cost
+        deployment_time
       }))
 {
   // Do nothing
@@ -156,12 +153,6 @@ const State& TaskPlanner::Assignment::state() const
 const rmf_traffic::Time& TaskPlanner::Assignment::deployment_time() const
 {
   return _pimpl->deployment_time;
-}
-
-//==============================================================================
-bool TaskPlanner::Assignment::include_in_cost() const
-{
-  return _pimpl->include_in_cost;
 }
 
 //==============================================================================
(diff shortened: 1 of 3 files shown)


Yadunund:
Could you commit a test case where the overall costs of the assignments is significantly improved?

mrushyendra:
Just added one

rmf_task/src/rmf_task/agv/TaskPlanner.cpp
```diff
@@ -742,44 +764,21 @@ class TaskPlanner::Implementation
...
-        // Clear out any infinity placeholders. Those candidates simply don't have
-        // any unassigned tasks that want to use it.
-        const auto& assignments = node.assigned_tasks[i];
-        if (assignments.empty())
-          value = 0.0;
-        else
-          value =
-            rmf_traffic::time::to_seconds(
-              assignments.back().state().finish_time() - time_now);
+        initial_queue_values[i] = rmf_traffic::time::to_seconds(
```
Yadunund, line 780: It seems that we are considering the finish times of charging tasks as well here while ignoring it in the `compute_g()` function. Should we be ignoring them here as well?

mrushyendra: Here we determine the earliest possible time the agent can begin any currently unassigned task. In the `InvariantHeuristicQueue` class, we use this value to determine the earliest time an unassigned task may finish (by adding the invariant costs leading up to it in the queue), then subtract from this finish time its specified earliest start time, mimicking `g(n)`. So we're only including the charging task when determining the earliest time the agent can begin a task, but not separately adding it to the queue (the initial queue value is a bit special in that sense).

mrushyendra:
I just pushed an additional change that incorporates more of the original solution. By also taking into account the smallest possible variant cost when determining the earliest time an agent can begin its next task, the heuristic should be slightly better.

Run times for the optimal solution decreased by as much as half, though still higher than the original version. Costs remain the same, as should be expected. For the greedy solution, run times are largely the same, and costs are slightly better on average (though not uniformly better).

rmf_task/src/rmf_task/agv/TaskPlanner.cpp
Yadunund: Ah right I overlooked that. Thanks for the clarification.

Yadunund (approved):
This is looking great! Thanks for improving the heuristic!

PR #208, opened 2020-11-06
Add unit test to detect implicit charging task at the start of Assignments

mrushyendra:
Added unit test that checks for discrepancy in state of charge to see if charging has been taken into account
without being explicitly added to `Assignments`.

### DIFF
--- a/rmf_task/test/unit/agv/test_TaskPlanner.cpp
+++ b/rmf_task/test/unit/agv/test_TaskPlanner.cpp
@@ -40,6 +40,10 @@
 
 #include <iostream>
 
+bool check_implicit_charging_task_start(
+  const rmf_task::agv::TaskPlanner::Assignments& assignments,
+  const double initial_soc);
+
 //==============================================================================
 inline void display_solution(
   std::string parent,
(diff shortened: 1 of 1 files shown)


rmf_task/test/unit/agv/test_TaskPlanner.cpp
```diff
@@ -359,4 +363,129 @@ SCENARIO("Grid World")
...
+    if (!agent.size())
+    {
+      continue;
+    }
+
+    const auto& s = agent[0].state();
+    // No task should consume more charge than (1.0 - initial_soc)
+    // in the current test, so we are guaranted to find any occurrence
+    // of an implicit charging task.
+    if (s.battery_soc() > initial_soc)
```
Yadunund, line 483: This condition should also check that agent[0].request() is not a ChargeBattery request(perhaps using `dynamic_cast`.  If not the test will fail even after we fix this problem.

rmf_task/test/unit/agv/test_TaskPlanner.cpp
```diff
@@ -40,6 +40,10 @@
 
 #include <iostream>
 
+bool check_implicit_charging_task_start(
```
Yadunund, line 43: Can we change this to an inline function with the definition moved here?

rmf_task/test/unit/agv/test_TaskPlanner.cpp
mrushyendra: Oops, had a check earlier to see if the new soc was close to 100%, but seem to have deleted it.

Yadunund: Lets have it check the request type? In the future we may support charging up the battery to levels other than 100%.

Yadunund (approved):
Thanks for writing up this test!

PR #209, opened 2020-11-06
Deal with implicit charging before first assignment

Yadunund:
When the initial battery soc of an agent is low and requires charging before proceeding with the first assignment, the planner implicitly assumes a charging task will be added while computing its best candidates. This PR keeps track of this assumption and appropriately appends a charging task to the list of assignments.

### DIFF
--- a/rmf_task/src/rmf_task/agv/TaskPlanner.cpp
+++ b/rmf_task/src/rmf_task/agv/TaskPlanner.cpp
@@ -185,6 +185,8 @@ public:
     std::size_t candidate;
     State state;
     rmf_traffic::Time wait_until;
+    State previous_state;
+    bool require_charge_battery = false;
   };
 
   // Map finish time to Entry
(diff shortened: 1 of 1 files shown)


rmf_task/src/rmf_task/agv/TaskPlanner.cpp
```diff
@@ -306,7 +315,12 @@ Candidates Candidates::make(
         assert(new_finish.has_value());
         initial_map.insert(
           {new_finish.value().finish_state().finish_time(),
-          Entry{i, new_finish.value().finish_state(), new_finish.value().wait_until()}});
+          Entry{
+            i,
+            new_finish.value().finish_state(),
+            new_finish.value().wait_until(),
```
mrushyendra, line 321: Here, the `wait_until` time of the entry with the charging task + original task is set to the time of the original task. I was wondering if that may be an issue in [this](https://github.com/osrf/rmf_core/blob/783c67a97c5e7220ec0a616d50b65b823513b2a1/rmf_task/src/rmf_task/agv/TaskPlanner.cpp#L884) line, where sometimes it might be more efficient to add a charging task much before we add the original task.

I think `expand_charger` should automatically take care of such cases, but I'm not a 100% sure.

mrushyendra (approved):
Just looked through all the changes, seems fine to me!

rmf_task/src/rmf_task/agv/TaskPlanner.cpp
Yadunund: Yeah I think `expand_charger()` should account for such cases.

PR #212, opened 2020-11-12
Cache estimated values to speed up TaskPlanner

mrushyendra:
Memoizes the computation of `variant_duration` and change in charge estimates between pairs of waypoints on the graph. Shortens run times for computing plans with the `TaskPlanner` by up to a couple of orders of magnitude, since it reduces the number of calls made to the `rmf_traffic::planner`, which formed the bulk of the run time previously.

### DIFF
--- a/rmf_task/include/rmf_task/Estimate.hpp
+++ b/rmf_task/include/rmf_task/Estimate.hpp
@@ -18,6 +18,9 @@
 #ifndef INCLUDE__RMF_TASK__ESTIMATE_HPP
 #define INCLUDE__RMF_TASK__ESTIMATE_HPP
 
+#include <optional>
+#include <utility>
+
 #include <rmf_task/agv/State.hpp>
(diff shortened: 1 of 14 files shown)


Yadunund (changes requested):
This is a fantastic improvement to the task planner!! Just some minor comments to address. Could we also modify the TaskPlanner tests such that the greedy and optimal solve methods are called in isolation? Right now the optimal solve method is called right after the greedy one and hence it leverages on the available cache to speed up the computation. As a result the computation times are heavily skewed. We could call these functions inside separate `WHEN()` blocks perhaps. That would isolate the caching.

mrushyendra:
I made the suggested changes. For the tests, I create a new TaskPlanner before calling each method, which should make the run times independent of each other while keeping the ability to compare both the optimal and greedy methods - I hope this is sufficient.

Yadunund (approved):
Thanks for addressing the issues!

PR #216, opened 2020-11-18
few liners to populate start and estimated finish time

tanyouliang95:
Populate start_time and estimated end_time in `task_summary` msg. Will soon need to address the issue of updating the estimated end_time when the task is delayed.

### DIFF
--- a/rmf_fleet_adapter/src/rmf_fleet_adapter/TaskManager.cpp
+++ b/rmf_fleet_adapter/src/rmf_fleet_adapter/TaskManager.cpp
@@ -155,10 +155,12 @@ void TaskManager::set_queue(
     if (i != 0)
       start = assignments[i-1].state().location();
     start.time(a.deployment_time());
+    rmf_task_msgs::msg::TaskType task_type_msg;
 
     if (const auto request =
       std::dynamic_pointer_cast<const rmf_task::requests::Clean>(a.request()))
     {
+      task_type_msg.type = task_type_msg.TYPE_CLEAN;
       auto task = rmf_fleet_adapter::tasks::make_clean(
         request,
         _context,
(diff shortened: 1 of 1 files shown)


rmf_fleet_adapter/src/rmf_fleet_adapter/TaskManager.cpp
```diff
@@ -175,6 +175,12 @@ void TaskManager::set_queue(
       msg.task_profile.task_id = _queue.back()->id();
       msg.state = msg.STATE_QUEUED;
       msg.robot_name = _context->name();
+      msg.task_profile.task_type.type =
```
Yadunund, line 178: Lets populate these fields for Loop and Delivery requests below as well

Yadunund (changes requested):
Thanks for populating these fields. One tiny comment to address before we merge

rmf_fleet_adapter/src/rmf_fleet_adapter/TaskManager.cpp
```diff
@@ -246,19 +217,25 @@ void TaskManager::set_queue(
...
-      msg.robot_name = _context->name();
-      this->_context->node()->task_summary()->publish(msg);
     }
 
     else
     {
       continue;
     }
+
+    // publish queued task
```
Yadunund, line 227: `if (publish_queue_msg)` then execute the block below.

Yadunund (changes requested):
It seems that with the new changes a TaskSummary message will get published with potential UB if an unsupported `request` is processed. ie even if a request is not successfully casted into a valid type. In these scenarios `task_type_msg` is uninitialized and `_queue.back()` may result in UB if `_queue` is empty.

Yadunund (approved):
The `continue` inside `else{}` should handle the scenarios with failed casts. My bad for overlooking this.

PR #217, opened 2020-11-23
Develop/dispatcher demo

Yadunund:
This PR introduces a two new libraries (`rmf_battery` & `rmf_task`) along with significant changes to the fleet adapters as summarized below:

1. `rmf_battery`
The objective of this library is to model the battery drain of mobile robots due to locomotion and on-board devices.
*  `rmf_battery::agv::BatterySystem` class which is a container for battery properties of the vehicle.
* `rmf_battery::agv::MechanicalSystem` class which is a container for mechanical properties of the vehicle.
* `rmf_battery::agv::PowerSystem` class which is a container for power system/device properties of the vehicle.
* `rmf_battery::MotionPowerSink` pure abstract class which contains `compute_change_in_charge(rmf_traffic::Trajectory trajectory)`. A sample implementation is provided by `rmf_battery::agv::SimpleMotionPowerSink`
*  `rmf_battery::DevicePowerSink` pure abstract class which contains `compute_change_in_charge(double run_time)`. A sample implementation is provided by `rmf_battery::agv::SimpleDevicePowerSink`

> The expectation is for the user to use a combination of `MotionPowerSink` and `DevicePowerSink(s)` to compute the expected battery drain for various applications 

2. `rmf_task`
The objective of this library is to define various task "requests" and a task allocation solver to assign `n` task requests across `m` robots while factoring in the battery drain for the robots. Charging tasks are automatically included in the plan such that robots will head back to their chargers to recharge before continuing with the remaining tasks.
* Factories for ChargeBattery, Clean, Deliver and Loop task requests (`rmf_task::requests::...`)
* `rmf_task::agv::TaskPlanner`: API for solving task allocation problem. Contains `greedy_solve()` and `optimal_solve()` methods. (With caching, the `optimal_solve()` method has comparable run times with the `greedy_solve()` method and produces solutions with lower costs)

3. Changes to `rmf_fleet_adapter`
* The soon-to-be deprecated `full_control` example updates the RMF schedule with the itinerary of waypoints in the docking path of a robot in docking mode
* Integration with dispatcher framework. `FleetUpdateHandle` subscribes to `rmf_task/bid_notice` topic to receive a new task request to accommodate into the robot's schedule. A proposal is computed and submitted via `rmf_task/bid_proposal` to the dispatcher node. When a `DispatchRequest` msg is received via `rmf_task/dispatch_request`, the schedule of the robot is updated with the previously computed proposal. 
* Added `set_task_planner_params()` , `set_recharge_threshold()`,  `accept_task_requests()` functions to FleetUpdateHandle API.
* Added `update_battery_soc()` and `get_participant()` methods to RobotUpdateHandle API.
* Various modifications to RobotContext, Task and TaskManager APIs. The TaskManager now has a timer which periodically checks if the next task in its queue is ready for deployment and if so will begin execution of the same.
* Added factories for converting `rmf_task::Request` into `rmf_fleet_adapter::Task` for various task requests (ChargeBattery, Clean, Delivery & Loop). These are found in `src/rmf_fleet_adapter/tasks/`
* `WaitForCharge` phase which finishes execution when the battery soc of the robot reaches/exceeds an expected percentage. 
* Updates to the full_control launch file to capture parameters required for battery modelling and task allocation planning.

4. `rmf_fleet_msgs`
* Added `Dock`, `DockParameter` and `DockSummary` msgs to propagate information on the docking paths for fleets. Presently, a `mock_docker` node publishes this information via a `dock_summary` topic which the fleet adapter subscribes to.

5. `rmf_task_msgs`
* Added `BidNotice`, `BidProposal`, `DispatchRequest`, `TaskProfile`, `TaskType` msgs which are used during the bidding process.

6. `rmf_task_ros2`
* Hosting a Task Dispatcher node, which facilitates the bidding process for all new task requests.
* Tracks all initiated tasks and provide api and service call interfaces for external system to interact with.

Summary of Dispatcher framework:
image
With the new system, we have a Dispatcher Node that initiates a bidding process for new task requests.  When the Dispatcher receives a new request from a UI, it sends out a BidNotice msg to all the fleet adapters. If a fleet adapter is able to process that request, it submits a BidProposal back to the Dispatcher. (The TaskPlanner is used by the fleet adapters to determine how best to accommodate the new request). The Dispatcher compares all the BidProposals received and then submits a DispatchRequest msg with the fleet name of the robot that the bid is awarded to. There are a couple different ways the Dispatcher evaluates the proposals such as fastest to finish, lowest cost,etc which can be configured.

Battery recharging is tightly integrated with the new task planner. ChargeBattery tasks are optimally injected into a robot's schedule when the robot has insufficient charge to fulfill a series of tasks. Currently we assume each robot in the map has a dedicated charging location as annotated with the `is_charger` option in the traffic editor map.


rmf_battery/include/rmf_battery/agv/BatterySystem.hpp
```diff
@@ -0,0 +1,155 @@
...
+  ///
+  /// \param[in] charging_current
+  ///   The rated current in A for charging the battery
+  ///
+  /// \param[in] type
+  ///   The chemisty type of the battery
+  ///
+  /// \param[in] profile
+  ///   The battery profile for this battery. This is only required for calling
+  ///   the get_voltage() method
```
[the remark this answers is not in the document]
mxgrey: I guess this is supposed to be `estimate_voltage()`.

rmf_battery/include/rmf_battery/agv/BatterySystem.hpp
```diff
@@ -0,0 +1,155 @@
...
+  BatterySystem(
+    double nominal_voltage,
+    double capacity,
+    double charging_current,
+    BatteryType type = BatteryType::LeadAcid,
+    rmf_utils::optional<BatteryProfile> profile = rmf_utils::nullopt);
+
+  /// Estimate the voltage of the battery given its state of charge. This function
+  /// will return a nullopt if the battery profile is not defined
+  rmf_utils::optional<double> estimate_voltage(const double soc) const;
```
mxgrey, line 124: Instead of doing this, I might suggest having a (non-member) function

```
double estimate_voltage(const BatterySystem& system, const BatteryProfile& profile);
```

Then this class doesn't have to store a `BatteryProfile`, assuming this function is the only reason to store one.

rmf_battery/include/rmf_battery/agv/MechanicalSystem.hpp
```diff
@@ -0,0 +1,59 @@
...
+namespace agv {
+
+
+class MechanicalSystem
+{
+public:
+
+  MechanicalSystem(
+    double mass,
+    double inertia,
```
mxgrey, line 33: I assume this is meant to be the moment of inertia about the z-axis. That should be made more explicit. Simply saying "inertia" usually refers to mass.

rmf_battery/include/rmf_battery/agv/MechanicalSystem.hpp
```diff
@@ -0,0 +1,59 @@
...
+    double friction_coefficient);
+
+  MechanicalSystem& mass(double mass);
+  double mass() const;
+
+  MechanicalSystem& inertia(double inertia);
+  double inertia() const;
+
+  MechanicalSystem& friction_coefficient(double friction_coeff);
+  double friction_coefficient() const;
```
mxgrey, line 43: What type of friction is this meant to be? Static friction? Kinetic friction?

What are we dragging that's creating friction? Is this meant to be friction in the rotation of a vehicle's wheels? Or just the general forces on the mechanical system that push against its motion?

For a user to know what kind of friction coefficient they provide, we should be explicit about what calculations we will use it for.

rmf_battery/include/rmf_battery/agv/PowerSystem.hpp
```diff
@@ -0,0 +1,63 @@
...
+  /// Constructor
+  ///
+  /// \param[in] name
+  ///   A string representing the name of this power system
+  ///
+  /// \param[in] nominal_power
+  ///   The rated nominal power consumption in Watts for this power system
+  ///
+  PowerSystem(
+    std::string name,
```
mxgrey, line 40: Why does a power system need to know its own name? Shouldn't the name be an external label that the power system itself doesn't need to care about?

mxgrey (commented):
I've started reviewing some of `rmf_battery`, but I've only looked through a few files so far. More reviews will be coming in as time permits.

rmf_battery/include/rmf_battery/agv/PowerSystem.hpp
Yadunund: I agree the `PowerSystem` should not require a name. This was previously needed for an older implementation that has changed since. Will get rid of it.

rmf_battery/include/rmf_battery/agv/MechanicalSystem.hpp
Yadunund: This is the kinetic friction in the rotation of the vehicles's wheels which is used to calculate the energy loss during locomotion. I will update the documentation to reflect this.

rmf_battery/include/rmf_battery/agv/MechanicalSystem.hpp
Yadunund: Yes this is the moment of inertia about the Z-Axis. Will update the documentation

rmf_battery/include/rmf_battery/agv/BatterySystem.hpp
Yadunund: Thanks for the suggestion. I'm considering getting rid of the `BatteryProfile` class along with `estimate_voltage()` function entirely. These are no longer being used in any downstream packages. We simply use the `state of charge` to keep track of the state of the battery.

rmf_battery/include/rmf_battery/agv/PowerSystem.hpp
Yadunund: Does it make more sense to get rid of the `PowerSystem` class and just supply the `nominal_power` to the constructor of `SimpleDevicePowerSinik` ?

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
```diff
@@ -61,217 +72,564 @@ class LiaisonNegotiator : public rmf_traffic::schedule::Negotiator
 } // anonymous namespace
 
 //==============================================================================
-auto FleetUpdateHandle::Implementation::estimate_delivery(
-    const rmf_task_msgs::msg::Delivery& request) const
--> rmf_utils::optional<FleetUpdateHandle::Implementation::DeliveryEstimate>
+void FleetUpdateHandle::Implementation::dock_summary_cb(
+  const DockSummary::SharedPtr msg)
```
mxgrey, line 76: Nitpick: This could be a `const DockSummary::SharedPtr&` to avoid unnecessary copying and reference counting.

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
```diff
@@ -61,217 +72,564 @@ class LiaisonNegotiator : public rmf_traffic::schedule::Negotiator
...
 
-  const auto loop_end_wp = graph.find_waypoint(request.finish_name);
-  if (!loop_end_wp)
-    return rmf_utils::nullopt;
+  // Process Cleaning task
+  if (task_type.type == rmf_task_msgs::msg::TaskType::TYPE_CLEAN)
+  {
+    if (task_profile.clean.start_waypoint.empty())
+    {
+      RCLCPP_INFO(
```
mxgrey, line 154: I would elevate this to at least `_WARN`, probably `_ERROR`, because missing a field like this indicates that something is broken somewhere in the upstream system.

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
```diff
@@ -61,217 +72,564 @@ class LiaisonNegotiator : public rmf_traffic::schedule::Negotiator
...
+        node->get_logger(),
+        "Dock param for dock_name:[%s] unavailable. Rejecting BidNotice with "
+        "task_id:[%s]", start_wp_name.c_str(), id.c_str());
+
+      return;
+    }
+    const auto clean_param = clean_param_it->second;
+
+    // Check for valid finish waypoint
+    const std::string finish_wp_name = clean_param.finish;
```
mxgrey, line 190: Nitpick: This could be `const std::string&` to avoid an unnecessary copy.

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
```diff
@@ -61,217 +72,564 @@ class LiaisonNegotiator : public rmf_traffic::schedule::Negotiator
...
+    if (clean_param_it == dock_param_map.end())
+    {
+      RCLCPP_INFO(
+        node->get_logger(),
+        "Dock param for dock_name:[%s] unavailable. Rejecting BidNotice with "
+        "task_id:[%s]", start_wp_name.c_str(), id.c_str());
+
+      return;
+    }
+    const auto clean_param = clean_param_it->second;
```
mxgrey, line 187: Nitpick: This could be `const auto&`.

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
```diff
@@ -61,217 +72,564 @@ class LiaisonNegotiator : public rmf_traffic::schedule::Negotiator
...
+    std::vector<Eigen::Vector3d>  positions;
+    for (const auto& location: clean_param.path)
+      positions.push_back({location.x, location.y, location.yaw});
+    rmf_traffic::Trajectory cleaning_trajectory =
+      rmf_traffic::agv::Interpolate::positions(
+        planner->get_configuration().vehicle_traits(),
+        start_time,
+        positions);
+    
+    if (!(cleaning_trajectory.size() > 0))
```
mxgrey, line 213: If you merge in the latest `master`, you can change this to `if (cleaning_trajectory.empty())`. Or even without the latest `master`, this could be `if (cleaning_trajectory.size() == 0)`

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
```diff
@@ -61,217 +72,564 @@ class LiaisonNegotiator : public rmf_traffic::schedule::Negotiator
...
+      node->get_logger(),
+      "Generated Clean request");
+  }
+
+  else if (task_type.type == rmf_task_msgs::msg::TaskType::TYPE_DELIVERY)
+  {
+    const auto& delivery = task_profile.delivery;
+    if (delivery.pickup_place_name.empty())
+    {
+      RCLCPP_INFO(
```
mxgrey, line 245: I'd recommend escalating this and the other `.empty()` tests to `_WARN` or `_ERROR`.

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
```diff
@@ -61,217 +72,564 @@ class LiaisonNegotiator : public rmf_traffic::schedule::Negotiator
...
+      motion_sink,
+      ambient_sink,
+      tool_sink,
+      planner,
+      start_time,
+      drain_battery);
+
+    RCLCPP_INFO(
+      node->get_logger(),
+      "Generated Clean request");
```
mxgrey, line 237: Maybe add the `task_id` to this message for better transparency.

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
```diff
@@ -61,217 +72,564 @@ class LiaisonNegotiator : public rmf_traffic::schedule::Negotiator
...
+      delivery.items,
+      motion_sink,
+      ambient_sink,
+      planner,
+      start_time,
+      drain_battery);
+
+    RCLCPP_INFO(
+      node->get_logger(),
+      "Generated Delivery request");
```
mxgrey, line 332: Maybe add the `task_id` to this message for better transparency.

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
```diff
@@ -61,217 +72,564 @@ class LiaisonNegotiator : public rmf_traffic::schedule::Negotiator
...
+
+  }
+  else if (task_type.type == rmf_task_msgs::msg::TaskType::TYPE_LOOP)
+  {
+    const auto& loop = task_profile.loop;
+    if (loop.start_name.empty())
+    {
+      RCLCPP_INFO(
```
mxgrey, line 340: I recommend escalating to `_WARN` or `_ERROR`, and same for `finish_name` below.

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
```diff
@@ -61,217 +72,564 @@ class LiaisonNegotiator : public rmf_traffic::schedule::Negotiator
...
+        "Rejecting BidNotice with task_id:[%s]" , id.c_str());
+
+      return;
+    }
+
+    if (loop.num_loops < 1)
+    {
+      RCLCPP_INFO(
+        node->get_logger(),
+        "Required param [loop.num_loops] in TaskProfile is invalid."
```
mxgrey, line 362: I'd suggest making this message more explicit, and perhaps escalating to `_WARN` or `_ERROR`.

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
```diff
@@ -61,217 +72,564 @@ class LiaisonNegotiator : public rmf_traffic::schedule::Negotiator
...
+      loop.num_loops,
+      motion_sink,
+      ambient_sink,
+      planner,
+      start_time,
+      drain_battery);
+
+    RCLCPP_INFO(
+      node->get_logger(),
+      "Generated Loop request");
```
mxgrey, line 405: Maybe add the `task_id` to this message for better transparency.

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
```diff
@@ -61,217 +72,564 @@ class LiaisonNegotiator : public rmf_traffic::schedule::Negotiator
...
+  }
+  else
+  {
+    RCLCPP_INFO(
+      node->get_logger(),
+      "Invalid TaskType in TaskProfile. Rejecting BidNotice with task_id:[%s]",
```
mxgrey, line 411: I'd suggest including the `task_type.type` information in this message and escalate it to `_WARN` or `_ERROR`.

mxgrey (commented):
Here's another batch of feedback. I still have a long way to go. I think I've only reviewed maybe half of this PR so far :sweat_smile:

rmf_battery/include/rmf_battery/agv/PowerSystem.hpp
mxgrey: If you don't anticipate the need to add more fields to the `PowerSystem` class in the future, then reducing it the way you've described sounds fine to me.

rmf_battery/include/rmf_battery/agv/MechanicalSystem.hpp
```diff
@@ -0,0 +1,59 @@
...
+
+  /// Returns true if the values are valid, i.e. greater than zero.
+  bool valid() const;
```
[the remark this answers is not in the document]
Yadunund: That makes a lot of sense. I have made the appropriate changes https://github.com/osrf/rmf_core/pull/217/commits/eb4ac7312e01824cf9381783455af4f128b065dd

rmf_battery/src/rmf_battery/agv/SimpleDevicePowerSink.cpp
```diff
@@ -0,0 +1,76 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+*/
+
+#include <rmf_battery/agv/SimpleDevicePowerSink.hpp>
+
+#include <cmath>
+
+#include <iostream>
```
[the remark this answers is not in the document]
Yadunund: Removed https://github.com/osrf/rmf_core/pull/217/commits/1b9af999e334db8fc9f67d4dd27cbd28c42276da

rmf_battery/src/rmf_battery/agv/SimpleMotionPowerSink.cpp
```diff
@@ -0,0 +1,130 @@
...
+
+#include <rmf_battery/agv/SimpleMotionPowerSink.hpp>
+
+#include <rmf_traffic/Motion.hpp>
+#include <rmf_traffic/Time.hpp>
+
+#include <vector>
+#include <cmath>
+
+#include <iostream>
```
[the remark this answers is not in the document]
Yadunund: Removed https://github.com/osrf/rmf_core/pull/217/commits/1b9af999e334db8fc9f67d4dd27cbd28c42276da

rmf_battery/src/rmf_battery/agv/SimpleMotionPowerSink.cpp
```diff
@@ -0,0 +1,130 @@
...
+  const double nominal_voltage = _pimpl->battery_system.nominal_voltage();
+  const double mass = _pimpl->mechanical_system.mass();
+  const double inertia = _pimpl->mechanical_system.inertia();
+  const double friction = _pimpl->mechanical_system.friction_coefficient();
+
+  auto begin_it = trajectory.begin();
+  auto end_it = --trajectory.end();
+
+  auto start_time = begin_it->time();
+  const auto end_time = end_it->time();
```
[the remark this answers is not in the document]
Yadunund: Done https://github.com/osrf/rmf_core/pull/217/commits/1b9af999e334db8fc9f67d4dd27cbd28c42276da

rmf_battery/src/rmf_battery/agv/SimpleMotionPowerSink.cpp
```diff
@@ -0,0 +1,130 @@
...
+
+} // namespace anonymous
+
+//==============================================================================
+double SimpleMotionPowerSink::compute_change_in_charge(
+  const rmf_traffic::Trajectory& trajectory) const
+{
+  assert(_pimpl->battery_system.valid());
+  assert(_pimpl->mechanical_system.valid());
+
```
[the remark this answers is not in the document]
Yadunund: Done https://github.com/osrf/rmf_core/pull/217/commits/1b9af999e334db8fc9f67d4dd27cbd28c42276da

rmf_battery/test/unit/agv/test_MechanicalSystem.cpp
```diff
@@ -0,0 +1,65 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+*/
+
+#include <rmf_battery/agv/MechanicalSystem.hpp>
+
+#include <rmf_utils/catch.hpp>
+
+
```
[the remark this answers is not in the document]
Yadunund: Removed https://github.com/osrf/rmf_core/pull/217/commits/1b9af999e334db8fc9f67d4dd27cbd28c42276da

rmf_battery/test/unit/agv/test_battery_drain.cpp
```diff
@@ -0,0 +1,296 @@
...
+
+
+#include <rmf_traffic/Trajectory.hpp>
+#include <rmf_traffic/agv/VehicleTraits.hpp>
+#include <rmf_traffic/agv/Interpolate.hpp>
+#include <rmf_traffic/Time.hpp>
+
+#include <rmf_utils/catch.hpp>
+
+#include <iostream>
```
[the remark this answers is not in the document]
Yadunund: Removed https://github.com/osrf/rmf_core/pull/217/commits/1b9af999e334db8fc9f67d4dd27cbd28c42276da

rmf_fleet_adapter/include/rmf_fleet_adapter/agv/FleetUpdateHandle.hpp
```diff
@@ -63,6 +68,63 @@ class FleetUpdateHandle : public std::enable_shared_from_this<FleetUpdateHandle>
...
+  ///   As a consequence, charging tasks will not be automatically assigned to
+  ///   vehicles in this fleet when battery levels fall below their thresholds.
+  ///
+  /// \return true if task planner parameters were successfully updated.
+  bool set_task_planner_params(
+    std::shared_ptr<rmf_battery::agv::BatterySystem> battery_system,
+    std::shared_ptr<rmf_battery::MotionPowerSink> motion_sink,
+    std::shared_ptr<rmf_battery::DevicePowerSink> ambient_sink,
+    std::shared_ptr<rmf_battery::DevicePowerSink> tool_sink,
+    const bool drain_battery);
```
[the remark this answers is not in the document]
Yadunund: Thanks for the suggestion. Added the function https://github.com/osrf/rmf_core/pull/217/commits/91d662409c24fbbe1aabe9c65a351df1bda21ec1

rmf_fleet_adapter/include/rmf_fleet_adapter/agv/FleetUpdateHandle.hpp
```diff
@@ -71,13 +133,17 @@ class FleetUpdateHandle : public std::enable_shared_from_this<FleetUpdateHandle>
...
+  /// \note This interface will be deprecated. Use the more general
+  ///   AcceptTaskRequest callback
   using AcceptDeliveryRequest =
       std::function<bool(const rmf_task_msgs::msg::Delivery& request)>;
 
   /// Provide a callback that indicates whether this fleet will accept a
   /// delivery request. By default all delivery requests will be rejected.
   ///
-  /// \note The callback function that you give should ideally be non-blocking
+  /// \note This function will be deprecated in favor of accept_task_requests()
```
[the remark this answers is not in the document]
Yadunund: Done https://github.com/osrf/rmf_core/pull/217/commits/91d662409c24fbbe1aabe9c65a351df1bda21ec1

rmf_battery/include/rmf_battery/agv/BatterySystem.hpp
```diff
@@ -0,0 +1,155 @@
...
+  ///   The chemisty type of the battery
+  ///
+  /// \param[in] profile
+  ///   The battery profile for this battery. This is only required for calling
+  ///   the get_voltage() method
+  BatterySystem(
+    double nominal_voltage,
+    double capacity,
+    double charging_current,
+    BatteryType type = BatteryType::LeadAcid,
```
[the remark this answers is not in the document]
Yadunund: Removed BatteryType https://github.com/osrf/rmf_core/pull/217/commits/eb4ac7312e01824cf9381783455af4f128b065dd

rmf_battery/include/rmf_battery/agv/BatterySystem.hpp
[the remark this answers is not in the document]
Yadunund: Removed this function definition https://github.com/osrf/rmf_core/pull/217/commits/eb4ac7312e01824cf9381783455af4f128b065dd

rmf_battery/include/rmf_battery/agv/BatterySystem.hpp
Yadunund: Removed BatteryProfile https://github.com/osrf/rmf_core/pull/217/commits/eb4ac7312e01824cf9381783455af4f128b065dd

rmf_battery/include/rmf_battery/agv/PowerSystem.hpp
Yadunund: Have decided to keep `PowerSystem` for now

rmf_battery/include/rmf_battery/agv/MechanicalSystem.hpp
Yadunund: https://github.com/osrf/rmf_core/pull/217/commits/d416f83a321ccf04a08fd1711248194babee5fbc

rmf_battery/include/rmf_battery/agv/MechanicalSystem.hpp
Yadunund: https://github.com/osrf/rmf_core/pull/217/commits/d416f83a321ccf04a08fd1711248194babee5fbc

rmf_fleet_adapter/include/rmf_fleet_adapter/agv/RobotUpdateHandle.hpp
```diff
@@ -97,6 +105,10 @@ class RobotUpdateHandle
   /// value that was given to the setter.
   rmf_utils::optional<rmf_traffic::Duration> maximum_delay() const;
 
+  /// Get the schedule participant of this robot
+  rmf_utils::optional<std::reference_wrapper<rmf_traffic::schedule::Participant>>
+    get_participant();
```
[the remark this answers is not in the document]
Yadunund: Agreed that giving access to the schedule participant handle is not appropriate in the long run. I have adopted the Unstable extension as suggested https://github.com/osrf/rmf_core/pull/217/commits/0688fa577df6ff13d2514c5c5acc6d7a99f4df3a

rmf_fleet_adapter/src/full_control/main.cpp
```diff
@@ -289,6 +333,11 @@ class FleetDriverRobotCommandHandle
       // command
       estimate_state(_node, state.location, _travel_info);
     }
+
+    // Update battery soc
+    const double battery_soc = state.battery_percent / 100.0;
+    if (battery_soc >= 0.0 && battery_soc <= 1.0)
```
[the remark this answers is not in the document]
Yadunund: Logged as error https://github.com/osrf/rmf_core/pull/217/commits/be64e8c4af5856af8f9c45c4e67a5a15698da165

rmf_fleet_adapter/src/rmf_fleet_adapter/Task.hpp
```diff
@@ -93,10 +97,15 @@ class Task : public std::enable_shared_from_this<Task>
   using PendingPhases = std::vector<std::unique_ptr<PendingPhase>>;
 
   // Make a new task
+  // TODO(YV) Remove default nullptr for request after refactoring Loop and
```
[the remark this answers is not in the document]
Yadunund: This has already been addressed. Removed the comment https://github.com/osrf/rmf_core/pull/217/commits/e78dd7938de085b4230390ee1c23ef79927bf309

rmf_fleet_adapter/src/rmf_fleet_adapter/TaskManager.cpp
```diff
@@ -17,6 +17,22 @@
...
+#include <rmf_traffic_ros2/Time.hpp>
+
+#include <rmf_traffic/agv/Planner.hpp>
+
+#include "tasks/Clean.hpp"
+#include "tasks/ChargeBattery.hpp"
+#include "tasks/Delivery.hpp"
+#include "tasks/Loop.hpp"
+
+#include <iostream>
```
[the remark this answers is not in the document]
Yadunund: Removed https://github.com/osrf/rmf_core/pull/217/commits/e78dd7938de085b4230390ee1c23ef79927bf309

rmf_fleet_adapter/src/rmf_fleet_adapter/TaskManager.cpp
```diff
@@ -38,6 +54,25 @@ TaskManagerPtr TaskManager::make(agv::RobotContextPtr context)
       }
     });
 
+  mgr->_task_timer = mgr->context()->node()->create_wall_timer(
```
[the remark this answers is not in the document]
Yadunund: Done https://github.com/osrf/rmf_core/issues/231

rmf_fleet_adapter/src/rmf_fleet_adapter/TaskManager.cpp
```diff
@@ -84,6 +121,26 @@ auto TaskManager::expected_finish_location() const -> StartSet
   return _context->location();
 }
 
+//==============================================================================
+auto TaskManager::expected_finish_state() const -> State
+{
+  // If an active task exists, return the estimated finish state of that task
+  /// else update the current time and battery level for the state and return
+  if (_active_task)
+    return _context->state();
```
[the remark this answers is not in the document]
Yadunund: Sorry for the confusion. I have renamed `RobotContext::state()` to `RobotContext::current_task_end_state()` as recommended. 
https://github.com/osrf/rmf_core/pull/217/commits/b8d61d65da5b67781eed5483e11ab80d650e70eb

rmf_fleet_adapter/src/rmf_fleet_adapter/TaskManager.cpp
```diff
@@ -96,67 +153,313 @@ agv::ConstRobotContextPtr TaskManager::context() const
...
+    return;
+  }
+
+  std::lock_guard<std::mutex> guard(_mutex);
+  const rmf_traffic::Time now = rmf_traffic_ros2::convert(
+    _context->node()->now());
+  const auto next_task = _queue.front();
+  const auto deployment_time = next_task->deployment_time();
+
+  if (now > deployment_time)
```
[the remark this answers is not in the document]
Yadunund: Done https://github.com/osrf/rmf_core/pull/217/commits/ece42e9badc59f55bbffda1e081657f9dce7d714

rmf_fleet_adapter/src/rmf_fleet_adapter/TaskManager.cpp
```diff
@@ -84,6 +121,26 @@ auto TaskManager::expected_finish_location() const -> StartSet
...
+//==============================================================================
+auto TaskManager::expected_finish_state() const -> State
+{
+  // If an active task exists, return the estimated finish state of that task
+  /// else update the current time and battery level for the state and return
+  if (_active_task)
+    return _context->state();
+
+  // Update battery soc and finish time in the current state
+  auto& finish_state = _context->state();
```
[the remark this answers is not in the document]
Yadunund: Done https://github.com/osrf/rmf_core/pull/217/commits/ece42e9badc59f55bbffda1e081657f9dce7d714

rmf_fleet_adapter/src/rmf_fleet_adapter/TaskManager.cpp
```diff
@@ -96,67 +153,313 @@ agv::ConstRobotContextPtr TaskManager::context() const
...
 
-  RCLCPP_INFO(
-        _context->node()->get_logger(),
-        "Beginning new task [%s] for [%s]. Remaining queue size: %d",
-        _active_task->id().c_str(), _context->requester_id().c_str(),
-        _queue.size());
+//==============================================================================
+void TaskManager::retreat_to_charger()
+{
+  if (_active_task || !_queue.empty())
```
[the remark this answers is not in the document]
Yadunund: Done https://github.com/osrf/rmf_core/pull/217/commits/81ee72b3599c5dc2dfd2a6e8b0171956babc555e

rmf_fleet_adapter/src/rmf_fleet_adapter/TaskManager.hpp
```diff
@@ -39,18 +43,37 @@ class TaskManager : public std::enable_shared_from_this<TaskManager>
...
 
+  /// Set the queue for this task manager with assignments generated from the
+  /// task planner
+  void set_queue(const std::vector<Assignment>& assignments);
+
+  /// Get the non-charging requests among pending tasks
+  const std::vector<rmf_task::ConstRequestPtr> requests() const;
+
+  /// Callback for task timer which begins next task if its deployment time has passed
+  void _begin_next_task();
```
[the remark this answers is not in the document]
Yadunund: Moved back into `private` scope https://github.com/osrf/rmf_core/pull/217/commits/d907d545c5f9f2b67dca882af25ffe71a6fddda2

rmf_fleet_adapter/src/rmf_fleet_adapter/TaskManager.cpp
```diff
@@ -96,67 +153,313 @@ agv::ConstRobotContextPtr TaskManager::context() const
...
+    // TODO(YV) Protect this call with a mutex
+    estimate_cache->set(endpoints, retreat_duration,
+      retreat_battery_drain);
+  }
+
+  const double battery_soc_after_retreat =
+    current_battery_soc - retreat_battery_drain;
+
+  if ((battery_soc_after_retreat < retreat_threshold) &&
+    (battery_soc_after_retreat > threshold_soc))
```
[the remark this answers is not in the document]
Yadunund: Print out a warning https://github.com/osrf/rmf_core/pull/217/commits/d0fcd46ce67b1739e2579b152caa584da836eb1f

rmf_fleet_adapter/src/rmf_fleet_adapter/TaskManager.cpp
```diff
@@ -96,67 +153,313 @@ agv::ConstRobotContextPtr TaskManager::context() const
...
-  },
-        [this, id = _active_task->id()]()
+    const double dSOC_motion =
+      task_planner_config->motion_sink()->compute_change_in_charge(trajectory);
+    const double dSOC_device =
+      task_planner_config->ambient_sink()->compute_change_in_charge(
+        rmf_traffic::time::to_seconds(retreat_duration));
+    retreat_battery_drain = dSOC_motion + dSOC_device;
+
+    // TODO(YV) Protect this call with a mutex
```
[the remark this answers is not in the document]
Yadunund: Protected read/write to EstimateCache with a mutex https://github.com/osrf/rmf_core/pull/217/commits/d0fcd46ce67b1739e2579b152caa584da836eb1f

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
```diff
@@ -61,217 +72,564 @@ class LiaisonNegotiator : public rmf_traffic::schedule::Negotiator
...
+      "[TaskPlanner] Failed to compute assignments for task_id:[%s]",
+      id.c_str());
+
+    return;
+  }
+
+  const double cost = task_planner->compute_cost(assignments);
+
+  // Display assignments for debugging
+  std::cout << "Cost: " << cost << std::endl;
```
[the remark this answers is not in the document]
Yadunund: I'm inclined to keep the printouts. Would that be okay?

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
Yadunund: Done https://github.com/osrf/rmf_core/pull/217/commits/f73f6896aedf9ba33247fe43c9034138912a848a

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
Yadunund: Done https://github.com/osrf/rmf_core/pull/217/commits/f73f6896aedf9ba33247fe43c9034138912a848a

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
Yadunund: Done https://github.com/osrf/rmf_core/pull/217/commits/f73f6896aedf9ba33247fe43c9034138912a848a

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
Yadunund: Done https://github.com/osrf/rmf_core/pull/217/commits/f73f6896aedf9ba33247fe43c9034138912a848a

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
Yadunund: Done https://github.com/osrf/rmf_core/pull/217/commits/f73f6896aedf9ba33247fe43c9034138912a848a

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
Yadunund: Done https://github.com/osrf/rmf_core/pull/217/commits/f73f6896aedf9ba33247fe43c9034138912a848a

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
Yadunund: Done https://github.com/osrf/rmf_core/pull/217/commits/f73f6896aedf9ba33247fe43c9034138912a848a

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
Yadunund: Done https://github.com/osrf/rmf_core/pull/217/commits/f73f6896aedf9ba33247fe43c9034138912a848a

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
Yadunund: Done https://github.com/osrf/rmf_core/pull/217/commits/f73f6896aedf9ba33247fe43c9034138912a848a

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
Yadunund: Done https://github.com/osrf/rmf_core/pull/217/commits/f73f6896aedf9ba33247fe43c9034138912a848a

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
Yadunund: Done https://github.com/osrf/rmf_core/pull/217/commits/f73f6896aedf9ba33247fe43c9034138912a848a

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
Yadunund: Done https://github.com/osrf/rmf_core/pull/217/commits/f73f6896aedf9ba33247fe43c9034138912a848a

rmf_fleet_adapter/src/rmf_fleet_adapter/TaskManager.cpp
```diff
@@ -96,67 +153,313 @@ agv::ConstRobotContextPtr TaskManager::context() const
...
+    const rmf_traffic::Duration retreat_duration =
+      finish_time - current_state.finish_time();
 
-    msg.task_id = id;
-    _context->node()->task_summary()->publish(msg);
-    _begin_next_task();
-  },
-        [this, id = _active_task->id()]()
+    const double dSOC_motion =
+      task_planner_config->motion_sink()->compute_change_in_charge(trajectory);
```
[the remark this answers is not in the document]
Yadunund: Thank you for pointing this out. I had actually overlooked this in `rmf_task` as well where we compute the battery drain for various tasks during task allocation planning. I have corrected the code to consider all the itineraries in a generated plan when determining battery drain. https://github.com/osrf/rmf_core/pull/217/commits/3e616c9e3e88da1bb33a95a151fc990ac00330dd

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/RobotContext.hpp
```diff
@@ -112,6 +116,30 @@ class RobotContext
       const TableViewerPtr& table_viewer,
       const ResponderPtr& responder) final;
 
+  /// Set the state of this robot at the end of its current task
+  RobotContext& current_task_end_state(const rmf_task::agv::State& state);
+
+  /// Get a mutable reference to the state of this robot at the end of its
+  // current task
+  rmf_task::agv::State& current_task_end_state();
```
mxgrey, line 124: It seems odd to provide this mutable access and also provide a setter function. This makes it impossible for a `const RobotContext&` to view this State field. I would recommend changing this function signature to

```
const rmf_task::agv::State& current_task_end_state() const;
```

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/test/MockAdapter.cpp
```diff
@@ -141,13 +141,13 @@ void MockAdapter::stop()
 //==============================================================================
 void MockAdapter::request_delivery(const rmf_task_msgs::msg::Delivery& request)
 {
-  rmf_fleet_adapter::agv::request_delivery(request, _pimpl->fleets);
+  // rmf_fleet_adapter::agv::request_delivery(request, _pimpl->fleets);
```
mxgrey, line 144: We should at least put in some TODOs to fix up these functions whose implementations are now dead. There may be tests that depend on them. Or if there aren't tests that depend on them, then the functions should be removed entirely.

rmf_fleet_adapter/src/rmf_fleet_adapter/load_param.cpp
```diff
@@ -88,4 +88,46 @@ rmf_traffic::agv::VehicleTraits get_traits_or_default(rclcpp::Node& node,
...
+std::optional<rmf_battery::agv::MechanicalSystem> get_mechanical_system(
+  rclcpp::Node& node,
+  const double default_mass,
+  const double default_inertia,
+  const double default_friction)
+{
+    const double mass =
+    get_parameter_or_default(node, "mass", default_mass);
+  const double inertia =
+    get_parameter_or_default(node, "inertia", default_inertia);
```
mxgrey, line 123: This `"inertia"` should also be changed to something more like `"moment"`.

mxgrey (commented):
Nice progress!

Here's another feedback dump. Some of it is a second iteration on the topics covered before.

rmf_battery/src/rmf_battery/agv/SimpleMotionPowerSink.cpp
```diff
@@ -0,0 +1,127 @@
...
+namespace {
+
+double compute_friction_energy(
+  const double f,
+  const double m,
+  const double v,
+  const double dt)
+{
+  const double g = 9.81; // ms-1
+  return f * m * g * v * dt;
```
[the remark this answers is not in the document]
Yadunund: The energy loss computed here is largely due to the resistance in the bearings of the wheels which prevents the robot from coasting indefinitely.

The rolling resistance frictional force can be expressed in a [similar manner as given by the sliding friction model](https://en.wikipedia.org/wiki/Rolling_resistance#Rolling_resistance_coefficient). Here, we first compute the frictional force `Fr = f*N = f*m*g` and then compute the work done over distance `v*dt`. The frictional energy loss due to rotating on the spot is not included as it will be negligible compared to the inertial energy loss. 

Another representation of the rolling resistance model is `Fr = b.N/r` where `r` is the radius of the wheel. Here `b` is another coefficient of friction per meter. Ultimately this can be reduced the model used above. 

Please let me know what you think.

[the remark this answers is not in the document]
mxgrey: I see, thanks for the reference!

From the description in the documentation, I would have expected that the `friction_coefficient` was referring to the "rolling resistance coefficient with dimension of length" or `b` in the link, so I was confused about this formula not having any radius.

After thinking about it more, I can understand how it would make sense to normalize the friction coefficient to exclude the mass, if we want to allow the robots to change their mass when payloads are added or removed.

My recommendation at this point would be to keep this formula the way it is and just add that link to the documentation for `friction_coefficient` and explicitly state that it refers to `Crr`, the dimensionless rolling resistance coefficient.

mxgrey (commented):
Some more feedback!

rmf_task_msgs/srv/CancelTask.srv
```diff
@@ -0,0 +1,12 @@
+# Submit Task | "POST" service call
```
mxgrey, line 1: This comment looks like a copy/paste error.

rmf_fleet_adapter/src/rmf_fleet_adapter/load_param.cpp
Yadunund: Renamed to `moment_of_inertia` https://github.com/osrf/rmf_core/pull/217/commits/6b661b051545cd2752059222760b1564da93004b

rmf_task_msgs/srv/CancelTask.srv
```diff
@@ -0,0 +1,12 @@
...
+
+# Whom that call this srv
```
mxgrey, line 3:
```suggestion
# Identifier for who is requesting the service
```

rmf_task_msgs/srv/GetTask.srv
```diff
@@ -0,0 +1,15 @@
+# Query list of submitted tasks | Get service call
+
+# Whom that call this srv
```
mxgrey, line 3:
```suggestion
# Identifier for who is requesting the service
```

rmf_task_msgs/srv/GetTask.srv
```diff
@@ -0,0 +1,15 @@
+# Query list of submitted tasks | Get service call
```
mxgrey, line 1: Maybe `GetTasks` or `GetTaskList` would be a more appropriate service name?

rmf_battery/src/rmf_battery/agv/SimpleDevicePowerSink.cpp
```diff
@@ -0,0 +1,71 @@
...
+double SimpleDevicePowerSink::compute_change_in_charge(
+  const double run_time) const
+{
+  const double capacity = _pimpl->battery_system.capacity();
+  const double nominal_voltage = _pimpl->battery_system.nominal_voltage();
+  const double nominal_power = _pimpl->power_system.nominal_power();
+
+  const double dE = nominal_power * run_time;
+  const double dQ = dE / nominal_voltage;
+  const double dSOC = dQ / (capacity * 3600.0);
```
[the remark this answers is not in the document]
Yadunund: Done https://github.com/osrf/rmf_core/pull/217/commits/ce7c2abcbbd9b296ab7e328cd225b9039640ab96

rmf_battery/test/unit/agv/test_battery_drain.cpp
```diff
@@ -0,0 +1,300 @@
...
+  using MechanicalSystem = rmf_battery::agv::MechanicalSystem;
+  using PowerSystem = rmf_battery::agv::PowerSystem;
+  using SimpleMotionPowerSink = rmf_battery::agv::SimpleMotionPowerSink;
+  using SimpleDevicePowerSink = rmf_battery::agv::SimpleDevicePowerSink;
+  using namespace std::chrono_literals;
+
+  // Initializing system traits
+  auto battery_system_optional = BatterySystem::make(12.0, 24.0, 2.0);
+  REQUIRE(battery_system_optional);
+  BatterySystem& battery_system = *battery_system_optional;
```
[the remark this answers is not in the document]
Yadunund: Thanks for catching this. Fixed https://github.com/osrf/rmf_core/pull/217/commits/ce7c2abcbbd9b296ab7e328cd225b9039640ab96

rmf_task_msgs/srv/GetTask.srv
```diff
@@ -0,0 +1,15 @@
...
+# Input the generated task ID during submission
+# if empty, provide all Submitted Tasks
+string[] task_id
+
+---
+
+bool success
+
+TaskSummary[] active_tasks
+TaskSummary[] terminated_tasks
```
mxgrey, line 15: How long is a fleet adapter supposed to remember its list of terminated tasks? We should make sure we don't have a memory pile-up for very long running processes.

rmf_battery/test/unit/agv/test_battery_drain.cpp
```diff
@@ -0,0 +1,300 @@
...
+    rmf_traffic::Trajectory trajectory =
+      rmf_traffic::agv::Interpolate::positions(traits, start_time, positions);
+
+    const double dSOC_motion = motion_power_sink.compute_change_in_charge(
+      trajectory);
+    const double dSOC_device = device_power_sink.compute_change_in_charge(
+      rmf_traffic::time::to_seconds(trajectory.duration()));
+
+    const double remaining_soc = initial_soc - dSOC_motion - dSOC_device;
+    const bool ok = remaining_soc > -1.0 && remaining_soc < 0.10;
```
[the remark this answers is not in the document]
Yadunund: Have added comments explaining the expected battery drain https://github.com/osrf/rmf_core/pull/217/commits/ce7c2abcbbd9b296ab7e328cd225b9039640ab96

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/RobotContext.hpp
Yadunund: Fixed https://github.com/osrf/rmf_core/pull/217/commits/51031ae5fcf45f5074c35aecdbb8df5ccc6f9976

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/test/MockAdapter.cpp
Yadunund: Added TODOs https://github.com/osrf/rmf_core/pull/217/commits/3beae9802bb5d8baa5559ef0bcef98fcc3cda408

rmf_task_msgs/srv/SubmitTask.srv
```diff
@@ -0,0 +1,32 @@
...
+# Task definitions
+Station station
+Loop loop
+Delivery delivery
+# Charge charge
+Clean clean
+# Patrol patrol
+
+# fleet selection evaluator
+uint8 evaluator
```
mxgrey, line 21: I'm not sure if it makes sense to have a field for specifying the evaluation type. To me this carries several implications that I don't think are necessarily true:
1. The evaluation method might change arbitrarily during runtime
2. All possible evaluation methods that any system integrator might want to use must comply with one of these types (i.e. system integrators are not free to come up with a custom evaluation type)
3. A single enum is sufficient to fully define the evaluation method

I think a better approach in the long-run would be defining an abstract interface for every place where we evaluate task allocation quality, and have some kind of modular or plugin approach to allow system integrators to specify what evaluation method they want to use.

That being said, we'll be putting this out as a WIP, so we don't need to change this approach right now. Perhaps we can just open a ticket for it.

mxgrey (commented):
I've made it through all the changes now. Here's my last big feedback dump :tada:

rmf_task/CMakeLists.txt
```diff
@@ -0,0 +1,143 @@
...
+
+add_library(rmf_task SHARED
+  ${lib_srcs}
+)
+
+target_link_libraries(rmf_task
+  PUBLIC
+    rmf_battery::rmf_battery
+  PRIVATE
+    ${rmf_dispenser_msgs_LIBRARIES}
```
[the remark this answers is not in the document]
Yadunund: Fixed https://github.com/osrf/rmf_core/pull/217/commits/db40f2abab10ae46a16ba96a9a706b0a0bb32546

rmf_task/CMakeLists.txt
```diff
@@ -0,0 +1,143 @@
...
+)
+
+target_include_directories(rmf_task
+  PUBLIC
+    $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
+    $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>
+    ${EIGEN3_INCLUDE_DIRS}
+  PRIVATE
+    ${rmf_dispenser_msgs_INCLUDE_DIRS}
```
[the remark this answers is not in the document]
Yadunund: Fixed https://github.com/osrf/rmf_core/pull/217/commits/db40f2abab10ae46a16ba96a9a706b0a0bb32546

rmf_task/include/rmf_task/Estimate.hpp
```diff
@@ -0,0 +1,91 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+*/
+
+#ifndef INCLUDE__RMF_TASK__ESTIMATE_HPP
```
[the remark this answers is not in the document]
Yadunund: Fixed https://github.com/osrf/rmf_core/pull/217/commits/db40f2abab10ae46a16ba96a9a706b0a0bb32546

rmf_task/include/rmf_task/Request.hpp
```diff
@@ -0,0 +1,63 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+*/
+
+#ifndef INCLUDE__RMF_TASK__TASK_HPP
```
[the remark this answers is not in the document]
Yadunund: Fixed https://github.com/osrf/rmf_core/pull/217/commits/db40f2abab10ae46a16ba96a9a706b0a0bb32546

rmf_task/include/rmf_task/agv/StateConfig.hpp
```diff
@@ -0,0 +1,52 @@
...
+
+  class Implementation;
+private:
+  rmf_utils::impl_ptr<Implementation> _pimpl;
+};
+
+} // namespace agv
+} // namespace rmf_task
+
+#endif // INCLUDE__RMF_TASK__AGV__STATE_HPP
```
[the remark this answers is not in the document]
Yadunund: Fixed https://github.com/osrf/rmf_core/pull/217/commits/db40f2abab10ae46a16ba96a9a706b0a0bb32546

rmf_task/include/rmf_task/agv/TaskPlanner.hpp
```diff
@@ -0,0 +1,211 @@
...
+
+#include <vector>
+#include <memory>
+#include <functional>
+#include <variant>
+
+namespace rmf_task {
+namespace agv {
+
+
```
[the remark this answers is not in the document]
Yadunund: Fixed https://github.com/osrf/rmf_core/pull/217/commits/db40f2abab10ae46a16ba96a9a706b0a0bb32546

rmf_task/include/rmf_task/agv/TaskPlanner.hpp
```diff
@@ -0,0 +1,211 @@
...
+      std::shared_ptr<rmf_battery::MotionPowerSink> motion_sink,
+      std::shared_ptr<rmf_battery::DevicePowerSink> ambient_sink,
+      std::shared_ptr<rmf_traffic::agv::Planner> planner,
+      FilterType filter_type= FilterType::Hash);
+
+    /// Get the battery system
+    rmf_battery::agv::BatterySystem& battery_system();
+
+    /// Set the battery_system
+    Configuration& battery_system(rmf_battery::agv::BatterySystem battery_system);
```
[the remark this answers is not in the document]
Yadunund: Whoops missed those out. https://github.com/osrf/rmf_core/pull/217/commits/4d28934176c28986170fa247488df67fbea3033f

rmf_task/include/rmf_task/agv/TaskPlanner.hpp
```diff
@@ -0,0 +1,211 @@
...
+
+
+
+//==============================================================================
+class TaskPlanner
+{
+public:
+
+  // The type of filter used for solving the task assignment problem
+  enum class FilterType
```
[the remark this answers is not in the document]
Yadunund: Removed from the public API. https://github.com/osrf/rmf_core/pull/217/commits/2ca073bd8071c65c5d55bb5fec5adc426ae69399

rmf_task/include/rmf_task/agv/StateConfig.hpp
```diff
@@ -0,0 +1,52 @@
...
+
+#ifndef INCLUDE__RMF_TASK__AGV__STATECONFIG_HPP
+#define INCLUDE__RMF_TASK__AGV__STATECONFIG_HPP
+
+#include <rmf_utils/impl_ptr.hpp>
+
+namespace rmf_task {
+namespace agv {
+
+class StateConfig
```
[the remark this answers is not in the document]
Yadunund: Ah yes that is a better way to look at the purpose of this class. `Constraints` may be a bit ambiguous especially when stored in the `RobotContext`. How about `StateConstraints`?

rmf_battery/src/rmf_battery/agv/SimpleMotionPowerSink.cpp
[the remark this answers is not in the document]
Yadunund: I've updated the documentation to reference the dimensionless coefficient. 
https://github.com/osrf/rmf_core/pull/217/commits/49fdf4469866a1ec13eedc2f295e0cb0fbcfd760

rmf_task/src/rmf_task/agv/TaskPlanner.cpp
```diff
@@ -0,0 +1,1440 @@
...
+{
+  if (_type == TaskPlanner::FilterType::Passthrough)
+    return false;
+
+  if (_type == TaskPlanner::FilterType::Hash)
+    return !_set.insert(node.assigned_tasks).second;
+
+  bool new_node = false;
+
+  // TODO(MXG): Consider replacing this tree structure with a hash set
```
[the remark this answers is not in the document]
Yadunund: Removed https://github.com/osrf/rmf_core/pull/217/commits/d3880be4738600ed8a62f9fd023732fcb6b670cc

rmf_task/include/rmf_task/requests/Loop.hpp
```diff
@@ -0,0 +1,90 @@
...
+
+  rmf_traffic::Time earliest_start_time() const final;
+
+  std::size_t start_waypoint() const;
+
+  std::size_t finish_waypoint() const;
+
+  std::size_t num_loops() const;
+
+  Start loop_start(const Start& start) const;
```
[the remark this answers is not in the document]
Yadunund: Added documentation https://github.com/osrf/rmf_core/pull/217/commits/d3880be4738600ed8a62f9fd023732fcb6b670cc

These functions were created in order to generate Starts for the different the `GoToPlace` phases for this task.
https://github.com/osrf/rmf_core/blob/0f57a1049b07dbd99c304bfc0cec0eb87d5e2588/rmf_fleet_adapter/src/rmf_fleet_adapter/tasks/Loop.cpp#L35-L55

rmf_task_msgs/srv/CancelTask.srv
Yadunund: Fixed https://github.com/osrf/rmf_core/commit/d3880be4738600ed8a62f9fd023732fcb6b670cc

rmf_task_msgs/srv/GetTask.srv
Yadunund: Renamed to `GetTaskList` https://github.com/osrf/rmf_core/pull/217/commits/0f57a1049b07dbd99c304bfc0cec0eb87d5e2588

@tanyouliang95 heads up to update the Dispatcher node with this new service name

rmf_task_msgs/srv/GetTask.srv
Yadunund: These service calls are meant to be made from UIs and are received by the Dispatcher node. Hence, the dispatcher node is the one storing a list of terminated tasks. I guess we need some kind of timeline for keeping track of such tasks. @tanyouliang95

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
```diff
@@ -297,6 +657,12 @@ void FleetUpdateHandle::add_robot(
          fleet = shared_from_this()](
         rmf_traffic::schedule::Participant participant)
   {
+    // TODO(YV) Get the battery % of this robot
+    const std::size_t charger = fleet->_pimpl->get_nearest_charger(
+        start[0], fleet->_pimpl->charging_waypoints);
+    rmf_task::agv::State state = rmf_task::agv::State{start[0], charger, 1.0};
```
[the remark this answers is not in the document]
Yadunund: Added `set_charger_waypoint(const std::size_t charger_wp)` method to `RobotUpdateHandle` which sets the intended charger for the robot. Would passing `const std::string charger_wp_name` with the name of the waypoint instead, enhance usability?
https://github.com/osrf/rmf_core/pull/217/commits/1488362e0921ee88a79e042369c311777419de04

rmf_task_msgs/srv/SubmitTask.srv
Yadunund: I agree with this. Have logged an issue to follow up with https://github.com/osrf/rmf_core/issues/237

rmf_task/src/rmf_task/Estimate.cpp
```diff
@@ -0,0 +1,115 @@
...
+
+//==============================================================================
+class EstimateCache::Implementation
+{
+  public:
+    struct PairHash
+    {
+      size_t operator()(const std::pair<size_t,size_t>& p) const
+      {
+        return std::hash<size_t>()(p.first) ^ std::hash<size_t>()(p.second);
```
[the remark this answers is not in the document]
Yadunund: Thanks for pointing out this shortcoming of using `xor` to combine hashes. I've made the change to use a sum based approach https://github.com/osrf/rmf_core/pull/217/commits/940b66faa98665e14fa12d59e88a9aa9fbd8c405

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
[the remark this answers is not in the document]
mxgrey: I think using the waypoint index is preferable. The user should always know how to lookup the waypoint index of a named waypoint. On the flip side, the waypoint of a charger might not necessarily be named (though admittedly that seems unlikely).

rmf_task/include/rmf_task/agv/StateConfig.hpp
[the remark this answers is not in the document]
mxgrey: The type name is already namespaced inside of `rmf_task::agv`, so I think that cuts down a lot on the ambiguity for the class name. For the instance that's stored inside of `RobotContext` you could give it a field name like `task_planning_constraints`.

mxgrey (commented):
Thanks for burning through all the feedback so quickly!

I just have one last follow-up that I think it worth addressing, shown below.

rmf_task/include/rmf_task/agv/StateConfig.hpp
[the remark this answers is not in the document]
Yadunund: Renamed to `Constraints` 
https://github.com/osrf/rmf_core/pull/217/commits/050be4f2f006a5b419d775b87007c1783062b78e

rmf_task/src/rmf_task/Estimate.cpp
```diff
@@ -0,0 +1,117 @@
...
+class EstimateCache::Implementation
+{
+  public:
+    struct PairHash
+    {
+      size_t operator()(const std::pair<size_t,size_t>& p) const
+      {
+        const auto hash_first = std::hash<size_t>()(p.first);
+        const auto hash_second = std::hash<size_t>()(p.second);
+        return  hash_first + hash_second + (hash_second << 10) ;
```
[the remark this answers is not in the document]
Yadunund: Thanks for pointing this out! I have implemented your suggestion https://github.com/osrf/rmf_core/pull/217/commits/29f4a28821b7240d0df08160b4d293bda60d257c

rmf_task_msgs/srv/GetTask.srv
mxgrey: I've opened an issue for this here: https://github.com/osrf/rmf_core/issues/238

mxgrey:
This all looks go to me. Let's just update the CHANGELOGs now to reflect the content of this PR.

I think the changes are sort of incomplete without https://github.com/osrf/rmf_core/pull/228, so I would suggest we get that to an approved state, then merge it into this PR, and then merge them both into `master` together.

rmf_task/src/rmf_task/Estimate.cpp
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    std::size_t _shift;
+  };
+
+  using Cache = std::unordered_map<std::pair<size_t,size_t>,
+    CacheElement, PairHash>;
+
+  Cache _cache;
+  std::shared_ptr<std::mutex> _mutex = std::make_shared<std::mutex>();
```
mxgrey, line 101: I think making this a `std::shared_ptr<std::mutex>` is solving the wrong problem.

My guess is that this choice was made because you were getting a compilation error from `rmf_utils::impl_ptr` about the copy constructor of `Implementation` being automatically deleted because the copy constructor of `std::mutex` is deleted. Switching to `std::shared_ptr<std::mutex>` seemed to magically make that compilation error disappear, and that's how we got here.

There's a problem here, though: This means every instance of `EstimateCache` that gets copied from one instance will share the same mutex, even though they're all storing data independently. The mutex will be "protecting" all of those instances from simultaneously read/writes even though the data of those instances is independent. This blocking will be completely unnecessary overhead.

Broadly speaking there are two options that I would recommend for making this all work correctly:

1. Use a `std::mutex` in `Implementation` and then use a `rmf_utils::unique_impl_ptr<Implementation>` for `EstimateCache` instead of `rmf_utils::impl_ptr<Implementation>`. This is the most straightforward solution if `EstimateCache` doesn't need to be copied.

2. Group together the `std::mutex` and `Cache` into their own struct and keep a `std::shared_ptr` of that whole struct. E.g.:

```
class EstimateCache::Implementation
{
  /* ... other stuff ... */

  struct Cache
  {
    using Data = std::unordered_map<std::pair<std::size_t, std::size_t>, CacheElement, PairHash>;
    Data data;
    std::mutex mutex;
  };

  std::shared_ptr<Cache> cache;
};
```

That way the mutex and the data are always tied together correctly. Copying may also be more efficient this way.

rmf_task/include/rmf_task/Estimate.hpp
```diff
@@ -0,0 +1,94 @@
...
+#include <utility>
+
+#include <rmf_task/agv/State.hpp>
+#include <rmf_traffic/Time.hpp>
+#include <rmf_utils/impl_ptr.hpp>
+
+namespace rmf_task {
+
+/// Estimates for requests
+class Estimate
```
mxgrey, line 31: I overlooked this concern in my first review, but is there any need for users to have access to this `Estimate` class? If it's only used internally by the task allocation planner, then there's no need for it to be in the public API. In that case, this header should be moved to `rmf_task/src/rmf_task/`. We should always try to minimize the surface area of the public API.

mxgrey (changes requested):
Two last things jumped out to me when doing the very last review pass-through, both of them related to the estimate cache.

rmf_task/src/rmf_task/Estimate.cpp
Yadunund: You're absolutely right about the reason behind using a shared ptr for the mutex. Since `EstimateCache` does not need to be copied, I've adopted the first approach. Thank you for highlighting this problem.
https://github.com/osrf/rmf_core/pull/217/commits/89a68c0f0bbc8a26994c302c37f9d275713d141a

rmf_task/include/rmf_task/Estimate.hpp
Yadunund: An `Estimate` object is returned by the `estimate_finish()` [method](https://github.com/osrf/rmf_core/blob/89a68c0f0bbc8a26994c302c37f9d275713d141a/rmf_task/include/rmf_task/Request.hpp#L44-L47) in `Request` and hence is exposed in the public API.

mxgrey: Ah, I see. How about `EstimateCache`? I'm not seeing where that's used by the public API.

Yadunund: `EstimateCache` is an argument in the `estimate_finish()` function. It is also returned by [this](https://github.com/osrf/rmf_core/blob/89a68c0f0bbc8a26994c302c37f9d275713d141a/rmf_task/include/rmf_task/agv/TaskPlanner.hpp#L192-L193) function in the `TaskPlanner` API which gets called in `TaskManager.cpp` [here](https://github.com/osrf/rmf_core/blob/89a68c0f0bbc8a26994c302c37f9d275713d141a/rmf_fleet_adapter/src/rmf_fleet_adapter/TaskManager.cpp#L391)

mxgrey: I'm somewhat unclear on the exact lifecycle of the `EstimateCache`, i.e. who is responsible for creating it and how it gets passed around to different objects. Generally I would expect any type of "cache" to be an implementation detail which is hidden from the public API. Ideally users shouldn't have to be aware that we're doing any kind of caching inside the library.

That being said, you've made it clear to me that it is currently a significant item in the public API, so we can move forward with it as-is and revisit this concern before the next release :+1:

mxgrey (approved):
Great job, everyone! :tada:

PR #218, opened 2020-11-25
Debug/task planner candidates

Yadunund:
Previously if a request is impossible to fulfil given battery constraints or improper value of recharge threshold, the task planner would throw an error. This PR addresses this behavior and instead returns an empty set of Assignments as the solution.

### DIFF
--- a/rmf_task/src/rmf_task/agv/TaskPlanner.cpp
+++ b/rmf_task/src/rmf_task/agv/TaskPlanner.cpp
@@ -193,7 +193,7 @@ public:
   // Map finish time to Entry
   using Map = std::multimap<rmf_traffic::Time, Entry>;
 
-  static Candidates make(
+  static std::shared_ptr<Candidates> make(
     const std::vector<State>& initial_states,
     const std::vector<StateConfig>& state_configs,
     const rmf_task::Request& request,
@@ -283,7 +283,7 @@ private:
   }
 };
 
-Candidates Candidates::make(
+std::shared_ptr<Candidates> Candidates::make(
   const std::vector<State>& initial_states,
   const std::vector<StateConfig>& state_configs,
   const rmf_task::Request& request,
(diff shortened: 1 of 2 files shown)


tanyouliang95 (approved):

mxgrey:
I'm not totally sure that returning an empty set is the best failure mode.

Ideally if a request is impossible to complete, the planner should communicate to the user why it was impossible, providing enough information for the user to consider taking some kind of remedial action (e.g. increase the recharge threshold).

What do you think the error handling process should be for a user who receives an empty assignment set?

Yadunund:
Hi @mxgrey 

I agree the that the planner should better communicate the reason behind not producing a plan. Would having a detailed message printout in addition to returning an empty set suffice? If not, what other mechanism do you recommend for conveying this information?

mxgrey:
Since this is meant to be a inner system library, printouts are not a good way of conveying information. Printouts require a human in the loop to read them, and we don't even know where our printout might land (maybe deep within a log somewhere that no one will ever see, or just casually discarded altogether).

I think the first thing to do here is figure out:
1. What are all the different possible failure modes (reasons for not being able to find a solution) of the planner?
2. What would be a reasonable response for a user to have for each failure mode?
3. Are there remedial actions that a user can take based on any of the failure modes to get a valid output on a second try?
4. Are there any failure modes where we know that nothing can be done to fix it?
5. How many different combinations of possible failure modes and potential responses exist?

One possibility is to use [`std::variant`](https://en.cppreference.com/w/cpp/utility/variant) to enumerate the different possible results of "success" + all the different failure modes (along with information about why the failure happened).

Another option is to return something similar to the [`rmf_traffic::agv::Planner::Result`](https://github.com/osrf/rmf_core/blob/42ba1f8173b96e24f40eb52280b47ba088c9b7e0/rmf_traffic/include/rmf_traffic/agv/Planner.hpp#L491) class, which represents the "result" of the planner, and which may have a successful state or an unsuccessful state. A Result instance with an unsuccessful value could contain information about why the planner was unsuccessful. A Result instance with a successful value could use `operator*` and `operator->` to dereference to the successful solution.

Yadunund:
Thank you for the comprehensive suggestions @mxgrey! I have begun to list out various failure modes and remedies as recommended along with examining the use to `using std::variant` or defining a `TaskPlanner::Result` class. Would it be okay to merge this in for the demo and open another PR for a more elegant fix?

mxgrey:
Sure, feel free to move things along as needed.

PR #219, opened 2020-11-30
add unit test for stubborn negotiator

ddengster:
The expected result of the negotiation with a StubbornNegotiator and SimpleNegotiator is for the negotiation object to be ready but incomplete.

### DIFF
--- a/rmf_traffic/test/unit/agv/test_Negotiator.cpp
+++ b/rmf_traffic/test/unit/agv/test_Negotiator.cpp
@@ -24,6 +24,7 @@
 #include <rmf_traffic/schedule/Database.hpp>
 #include <rmf_traffic/schedule/Negotiation.hpp>
 #include <rmf_traffic/schedule/Participant.hpp>
+#include <rmf_traffic/schedule/StubbornNegotiator.hpp>
 
 //==============================================================================
 void print_proposal(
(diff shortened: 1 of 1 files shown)


rmf_traffic/test/unit/agv/test_Negotiator.cpp
```diff
@@ -204,59 +205,97 @@ SCENARIO("Test Plan Negotiation Between Two Participants")
...
+      CHECK(negotiation->ready());
+
+      next_table = negotiation->table(p2.id(), {p1.id()});
+      // there should be a rejection of the alternative itinerary
+      negotiator_2.respond(
+        next_table->viewer(),
+        rmf_traffic::schedule::SimpleResponder::make(next_table));
+
+      // in this scenario, the negotiation never completes
+      CHECK_FALSE(negotiation->complete());
```
mxgrey, line 297: These expectations are all good, but after this `CHECK_FALSE` you could continue the negotiation process to completion.

After the `negotiator_2` rejects table `[0]`, you can give table `[0]` back to `negotiator_1`, and it should be able to give a proposal that accommodates `negotiator_2`. At that point, giving table `[0, 1]` to `negotiator_2` should succeed, and then the negotiation should be complete.

mxgrey: Note that what I described works because when `negotiator_2` rejects table `[0]` it also leaves a hint for `negotiator_1` about how `negotiator_1` can plan around it. Providing that hint is the key difference between "reject" and "forfeit".

ddengster: >  At that point, giving table [0, 1] to negotiator_2 should succeed, and then the negotiation should be complete.

(1) If I'm doing this right, adding the following after `CHECK_FALSE(negotiation->complete());` yields an incomplete solution despite it not going down the StubbornNegotiator::respond rejection path.
```
next_table = negotiation->table(p1.id(), {p2.id()});
negotiator_2.respond(
        next_table->viewer(),
        rmf_traffic::schedule::SimpleResponder::make(next_table)); //not rejected

CHECK(negotiation->complete()); //fails
```

(2) whereas the following succeeds. Correct me if I'm wrong, but it seems negotiator_2 has accepted the original p1's proposals.
```
next_table = negotiation->table(p1.id(), {});
negotiator_2.respond(
    next_table->viewer(),
    rmf_traffic::schedule::SimpleResponder::make(next_table));
CHECK(negotiation->complete()); //succeeds
```

This code crashes, and I suppose there's no table `[1, 0]` it's rejected.
```
next_table = negotiation->table(p2.id(), {p1.id()});
```

Let me know if anything's wrong with my assumptions. I'm putting (2) into the commit.

mxgrey: What you should add after `CHECK_FALSE(negotiation->complete());` is the following:

```
next_table = negotiation->table(p1.id(), {});
negotiator_1.respond(
  next_table->viewer(),
  rmf_traffic::schedule::SimpleResponder::make(next_table));

// TODO: Test that negotiator_1 found a solution

next_table = negotiation->table(p2.id(), {p1.id()});
negotiator_2.respond(
  next_table->viewer(),
  rmf_traffic::schedule::SimpleResponder::make(next_table));

// TODO: Test that the negotiation is complete
```

You need to have `negotiator_1` respond again. In the code that you've pushed, you're having `negotiator_2` respond to a table that belongs to `negotiator_1`.

ddengster: This LOC `next_table = negotiation->table(p2.id(), {p1.id()});` returns a null pointer.

mxgrey: That's problematic. Please push your latest version of the test with the failure happening, and I'll take a closer look at it. You can use `REQUIRE(next_table);` to make the test fail gracefully.

ddengster: 9d99ce6

rmf_traffic/test/unit/agv/test_Negotiator.cpp
```diff
@@ -204,59 +205,113 @@ SCENARIO("Test Plan Negotiation Between Two Participants")
...
+      // there should be a rejection of the proposed itinerary
+      negotiator_2.respond(
+        next_table->viewer(),
+        rmf_traffic::schedule::SimpleResponder::make(next_table));
+
+      // at this point, the negotiation is incomplete
+      CHECK_FALSE(negotiation->complete());
+
+      next_table = negotiation->table(p1.id(), {});
+      negotiator_2.respond(
```
mxgrey, line 300: This is still using `negotiator_2` when it needs to be calling `negotiator_1` here.

The idea here is that `negotiator_1` needs to re-respond after `negotiator_2`'s rejection, because now `negotiator_1` has more information that it can use when making its plan.

rmf_traffic/test/unit/agv/test_Negotiator.cpp
```diff
@@ -204,59 +205,113 @@ SCENARIO("Test Plan Negotiation Between Two Participants")
...
+        next_table->viewer(),
+        rmf_traffic::schedule::SimpleResponder::make(next_table));
+
+      CHECK(negotiation->complete());
```
mxgrey, line 304: The negotiation should not be complete yet, because now `negotiator_2` needs to re-respond before it's complete. This should be `CHECK_FALSE(negotiation->complete());`

rmf_traffic/test/unit/agv/test_Negotiator.cpp
ddengster: alright, that fixed it. thanks! 03b955f

mxgrey (approved):
Thanks for the additional test!

PR #221, opened 2020-11-30
TaskPlanner returns an std::variant

Yadunund:
This PR aims to address the issue of better conveying the reason behind the task planner failing to find a solution as brought up in (https://github.com/osrf/rmf_core/pull/218) 

The `optimal_solve()` and `greedy_solve()` methods in `TaskPlanner` now return a `Result`, which is an alias for `std::variant<TaskPlanner::Assignments, TaskPlanner::TaskPlannerError`>

`TaskPlannerError` is a `enum` class which has two members
* `low_battery`:  When none of the agents in the initial states have sufficient initial charge to even head back to their charging stations. Manual intervention is needed to recharge one or more agents.
* `limited_capacity`: When none of the agents in the initial states have sufficient battery capacity to accommodate one or more requests. This may be remedied by increasing the battery capacity or by lowering the threshold_soc in the state configs or by modifying the original request.


tanyouliang95 (approved):
Tested Task with extended distance, works well with expected `limited_capacity` error Result. :+1: . One helpful change is to use `RCLCPP_WARN/ERROR` in the current logging.

mxgrey (approved):

PR #223, opened 2020-12-02
Take into account battery drain while idling in TaskPlanner estimates

mrushyendra:
Modifies `estimate_finish` methods for `Delivery`, `Loop` and `Clean` requests to take into account battery drain while waiting to move towards the start waypoint for the task. Robots are assumed to have zero battery drain while waiting at a charger way point. In the future I assume this can be extended to calculate a potential increase in charge as well. Lastly, it also renames members of `Delivery::Implementation`.


Yadunund (commented):
This looks good! Can you also remove the leading underscore from the member variables in `ChargeBattery::Implementation`?

Yadunund (approved):

PR #226, opened 2020-12-04
Easy Traffic Light

mxgrey:
This PR introduces the `EasyTrafficLight` API. This new API is much simpler than the original `TrafficLight` API (which will be getting some breaking changes from this PR; the breaking changes were necessary in order to interoperate with the full control fleet adapters).

This also massively overhauls the implementation of the traffic light system in general. There is now a moderator whose job is to guarantee that traffic light robots never get into a permanent deadlock (as long as some very basic conditions are met by the traffic light fleet adapter user).

This PR depends on PRs the following two repos:
* https://github.com/osrf/rmf_demos/pull/168
* https://github.com/osrf/traffic_editor/pull/267

There are a few known issues remaining, but I don't consider these issues to be critical enough to block these PRs from moving forward. They can be address in later iterations:

1. Some kind of issue with how we're using `rxcpp` is causing the traffic light fleet adapter to get temporarily hung when difficult negotiations are going on. When this happens, it will look as though the fleet adapter has died (e.g. all of the traffic light robots will come to a stop and the schedule will stop updating for them), but really it just needs 15-20 seconds to resume normal operation. This should be totally fixable by auditing the way we're using `rxcpp`, but I'm going to leave that for a later fix.

2. There's a timing estimation issue when traffic light robots are turning corners which results in the message ` Failed to compute timing estimate for [tinyRobot2] owned by [tinyRobot2] moving away from checkpoint index [4]`. This should be a relatively simple math issue to fix, but it hasn't been a showstopper so far, so I'm going to leave it for a later PR.

3. In the office test (described below), the full control robot tends to get stuck on a chair because it cuts a corner too deeply (picture provided below). This is because of known (semi-intentional) behavior in the full control fleet adapter. In the long term this can be fixed by exposing a parameter for the amount of corner-cutting that we permit, but I'll save that for a future PR. In the short term, when this happens in a simulation, it can be fixed by manually dragging the robot away from the desk using Gazebo's translation tool.
world_gzclient_camera(1)-2020-12-04T16_10_28 921128

4. When mixing traffic light fleet adapter and full control fleet adapter in the same environment, it seems that a rare situation can arise where the traffic light robot's schedule vanishes and does not get updated, which can lead to a permanent deadlock. I'll be investigating this further in a follow-up PR.

There are three scenarios in the `rmf_demos` PR that are helpful for testing the traffic light fleet adapter:

Triple-H scenario:
```
$ ros2 launch demos triple_H.launch.xml
(new terminal) $ ros2 launch demos the_pedigree.launch.xml
```

Battle Royale Scenario:
```
$ ros2 launch demos battle_royale.launch.xml
(new terminal) $ ros2 launch demos battle_go.launch.xml
```

Office Scenario:
```
$ ros2 launch demos office_mock_traffic_light.launch.xml
(new terminal) $ ros2 launch demos office_traffic_light_test.launch.xml
```

The key thing to watch when running these tests is that they never permanently deadlock. If it seems like the robots are stuck, make sure it's not issue (1) by simply giving them time to unfreeze (if it lasts longer than a minute, then that's concerning), and make sure it's not issue (2) by seeing if the full control robot has gotten snagged on some office furniture. If it's issue (4), then you can manually intervene by moving the full control robot out of the way of the traffic light robot.


Yadunund (approved):
Did not run into any problems during my testing :sparkles:

rmf_fleet_adapter/include/rmf_fleet_adapter/agv/Adapter.hpp
```diff
@@ -102,15 +103,76 @@ class Adapter : public std::enable_shared_from_this<Adapter>
...
+  /// that can only support pause and resume commands.
+  ///
+  /// This API is much simpler to use than the standard traffic light API, but
+  /// it provides less information about the exact timing needed for the starts
+  /// and stops.
+  ///
+  /// This API should only be used for demo purposes, or if system integrators
+  /// can ensure very low-latency and reliable connections to the robots to
+  /// ensure that the commands arrive on time.
+  ///
```
tanyouliang95, line 142: missing description for `\handle_callback` param

rmf_fleet_adapter/include/rmf_fleet_adapter/agv/TrafficLight.hpp
```diff
@@ -112,8 +126,8 @@ class TrafficLight
     /// must not depart from a waypoint before it receives checkpoint
     /// information for that waypoint.
     ///
-    /// The next sequence of checkpoints will not be given until the on_standby
-    /// callback gets triggered.
+    /// The last checkpoint in the batch is the last checkpoint that the robot
+    /// is allowed to depart from. New batches override previous batches, so
```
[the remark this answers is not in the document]
mxgrey: This dangling sentence should be completed now.

rmf_fleet_adapter/include/rmf_fleet_adapter/agv/Adapter.hpp
mxgrey: Added!

PR #230, opened 2020-12-14
add Tow message

codebot:
Prototype of a Tow task, where a robot docks with "something" (a cart, or a shelving unit, etc.) and moves it somewhere else.

Signed-off-by: Morgan Quigley <morgan@osrfoundation.org>


mxgrey (approved):
:ok_hand:

PR #232, opened 2020-12-16
Give fleet adapters the option to publish fleet states (on by default)

mxgrey:


mxgrey:
This has been tested in simulation and on hardware, and appears to be working as intended.

PR #233, opened 2020-12-17
Add NegotiationCancel Message

arjo129:
Add a negotiation cancel message to cancel current negotiation.

GUI available: https://github.com/osrf/rmf_schedule_visualizer/pull/84


mxgrey:
I'm a bit embarrassed to say that when I put you on this task, I forgot that there is already a `NegotiationRefusal.msg` that already perfectly covers the functionality of this negotiation canceling.

I'm going to close this PR since it's redundant with an existing feature, and I'll update the schedule visualizer PR to use the already existing topic+message.

PR #234, opened 2020-12-17
removing fcl as its not a rosdep key

marcoag:
`fcl` is currently not a rosdep key therefore it does not resolve to any package. I assume it's trying to install `libfcl0.5` but this is already a dependency on `libfcl-dev` so I would suggest leaving this out of the `package.xml` for now.

Signed-off-by: Marco A. Gutierrez <marco@openrobotics.org>

### DIFF
--- a/rmf_traffic/package.xml
+++ b/rmf_traffic/package.xml
@@ -13,6 +13,8 @@
 
   <build_depend>libccd-dev</build_depend>
   <build_depend>libfcl-dev</build_depend>
+  <!-- Uncomment the line below for fcl-0.6 -->
+  <!-- <build_depend>fcl</build_depend> -->
 
   <test_depend>ament_cmake_catch2</test_depend>
   <test_depend>rmf_cmake_uncrustify</test_depend>
--- a/rmf_traffic/package.xml
+++ b/rmf_traffic/package.xml
@@ -13,7 +13,6 @@
 
   <build_depend>libccd-dev</build_depend>
   <build_depend>libfcl-dev</build_depend>
-  <build_depend>fcl</build_depend>
 
   <test_depend>ament_cmake_catch2</test_depend>
   <test_depend>rmf_cmake_uncrustify</test_depend>


mxgrey:
We actually do need it for fcl-0.6 support, because [in version 0.6 fcl comes with a `package.xml`](https://github.com/flexible-collision-library/fcl/blob/da07e5b4b59a99fcc45ea17a7aa786d5e077fdc9/package.xml#L2) where it declares itself as `fcl`.

That version of fcl isn't in ROS yet, but I expect it will be relatively soon, and this `<build_depend>` is needed to build against it from source using colcon.

I think it's best to keep it in unless it's causing some specific problem with build farm.

marcoag:
It is causing problems in the build farm as the rosdep key does not have a package to resolve to and the dependencies for the package cannot be created. 

`rosdep install --from-paths src --ignore-src -yr` will also fail with this `package.xml`

I could create a patch for this if they key is needed for source compilation with fcl-0.6 but that only solves the buildfarm side.

Alternatively I could propose a rosdep key for `fcl` that points to `libfcl0.5`, that way we would solve the `rosdep install` issue too.

mxgrey:
> Alternatively I could propose a rosdep key for fcl that points to libfcl0.5, that way we would solve the rosdep install issue too.

This sounds like the ideal solution to me. It would probably help many others with this transition period as well.

marcoag:
Suggested here: https://github.com/ros/rosdistro/pull/27789

ddengster:
Just a note that this change breaks the compile to fcl 0.6; we should let the rosdistro PR get resolved as suggested by @marcoag.

Until then, for any other observers who want to compile with fcl 0.6:

Just remove these 2 lines
```
  <build_depend>libccd-dev</build_depend>
  <build_depend>libfcl-dev</build_depend>
```
and add this tag:
```
  <build_depend>fcl</build_depend>
```

mxgrey:
I've added a comment in the `package.xml` for users who want fcl-0.6. It's not the best way to resolve this issue, but I think the better solutions are blocked for now.

mxgrey (approved):

PR #236, opened 2020-12-18
loop delivery tests for dispatcher

tanyouliang95:
Update previously breaking loop and delivery tasks test, related to the task dispatching system:

### To Run
Loop Test
```bash
./build/rmf_fleet_adapter/rmf_fleet_adapter_test "Scenario: Test loop requests"
````

Delivery Test
```bash
./build/rmf_fleet_adapter/rmf_fleet_adapter_test "Scenario: Test Delivery"
```

### DIFF
--- a/rmf_fleet_adapter/include/rmf_fleet_adapter/agv/test/MockAdapter.hpp
+++ b/rmf_fleet_adapter/include/rmf_fleet_adapter/agv/test/MockAdapter.hpp
@@ -22,8 +22,7 @@
 
 #include <rclcpp/node.hpp>
 
-#include <rmf_task_msgs/msg/delivery.hpp>
-#include <rmf_task_msgs/msg/loop.hpp>
+#include <rmf_task_msgs/msg/task_profile.hpp>
 
 #include <rmf_traffic/schedule/Snapshot.hpp>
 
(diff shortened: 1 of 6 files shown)


Yadunund (commented):
Thanks for getting started on fixing these tests!

Some high-level feedback:
I do not think we should add a `dispatch_task()` method to `FleetUpdateHandle`. It is never invoked in the implementation of the fleet adapter and hence not used during the actual bidding process. Only the MockAdapter seems to call this function which makes its usage not reflective of the actual processing done by the fleet adapters.

I think a better approach to testing the `Loop` and `Delivery` tasks could be to first create a `MockDispatcher` object which can be used to call a `dispatch_task(TaskProfile& task_profile, std::string& fleet_name)` within the `test_Loop` and `test_Delivery` scripts resp. Here the `fleet_name` can be the name of the fleet to award the job to. 

Let me know what you think.

tanyouliang95:
The reason for adding a `dispatch_task()` internal method is similar to the previous `request_delivery` and `request_loop` methods, defined [here](https://github.com/osrf/rmf_core/blob/0bf4ebbdd91141784f90b54f70ed113c8b1aa50b/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/internal_FleetUpdateHandle.hpp#L196-L203). This internal method is only used by `MockAdapter` for unit test of the fleet adapter.

Surely, from what you proposed, I do agree with the value of having an additional "Integration test", specific for both fleet adapter and dispatcher_node. I can add this separate test script in a separate PR. Ultimately this update is to address the depreciation of the previous task request method.

Yadunund:
But there is a critical difference as these functions were not defined solely for running the test; The previous `request_delivery()` and `request_loop()` functions were being called in `Adapter.cpp` as seen [here](https://github.com/osrf/rmf_core/blob/0bf4ebbdd91141784f90b54f70ed113c8b1aa50b/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/Adapter.cpp#L118) and [here](https://github.com/osrf/rmf_core/blob/0bf4ebbdd91141784f90b54f70ed113c8b1aa50b/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/Adapter.cpp#L132). Hence they were part of the 

However in the proposed changes we are adding a new function into the `FleetUpdateHandle` source file that will only be used in the test.

mxgrey:
Since `dispatch_task()` is only being used by `MockAdapter`, I'd suggest just moving the content of `dispatch_task()` into `MockAdapter::dispatch_task()` and deleting `dispatch_task()`.

None of this is in the public API, so we're free to refactor it however we please. Since `FleetUpdateHandle` isn't making use of `dispatch_task()`, it's not very self-consistent to have `FleetUpdateHandle.cpp` define that function. If something ever changes in the future where `FleetUpdateHandle` does start using it, we can refactor back to this approach when that time comes.

tanyouliang95:
Thanks for the feedbacks. This makes sense. I have removed the unnecessary function in `FleetUpdateHandle`, and the impl is in mockAdapter.

Yadunund (approved):

PR #246, opened 2021-01-04
fix rmf_fleet_adapter cmake to be used by packages downstream

ddengster:
allows `rmf_fleet_adapter::rmf_fleet_adapter` to be used in cmake's `target_include_directories` and `target_link_libraries`


mxgrey:
I'm not very familiar with these specific ament functions. Is `ament_export_targets` not sufficient for some reason? Other downstream packages (like `rmf_fleet_adapter_python`) seem to build against the imported `rmf_fleet_adapter::rmf_fleet_adapter` target without any problem. Are there some specific cmake configuration errors that you're getting without these ament function calls?

ddengster:
It seems that a couple of software updates to my system have been screwing up my builds, i tried building on my system without this PR and it works. So, I'll close it this PR

PR #261, opened 2021-01-15
Fix iterator invalidation

mxgrey:
An iterator invalidation bug was producing undefined behavior that may result in seg faults. This PR fixes that issue, and also silences some test failures that are being caused by bugs in FCL0.5 (the silenced tests are targeting unsupported features of `rmf_traffic`, so they do not have to actually be fixed).


ddengster (approved):
Thanks for the PR. `test_rmf_traffic` is running sucessfully, changes LGTM.

PR #262, opened 2021-01-15
Fix the cmake of rmf_fleet_adapter

mxgrey:
`rmf_task` a new dependency of `rmf_fleet_adapter`, but it was not being exported by ament. This causes linker errors for downstream packages that depend on `rmf_fleet_adapter`.


Yadunund (approved):
Thank you for addressing this 🙇🏼

PR #263, opened 2021-01-15
Add battery updating to the traffic light

mxgrey:


rmf_fleet_adapter/src/rmf_fleet_adapter/agv/EasyTrafficLight.cpp
```diff
@@ -552,6 +552,13 @@ std::size_t EasyTrafficLight::last_reached() const
   return _pimpl->last_reached;
 }
 
+//==============================================================================
+EasyTrafficLight& EasyTrafficLight::update_battery_soc(double battery_soc)
+{
+  _pimpl->update_handle->update_battery_soc(battery_soc);
```
Yadunund, line 558: Should we be calling `_pimpl->update_handle->update_battery_soc(battery_soc);` only `if (battery_soc >= 0.0 && battery_soc <= 1.0)` ?

mxgrey: Right now updating the battery SOC for the traffic light APIs will do nothing but forward that value into the robot state messages that get published (which then only goes to human operators), so there won't really be any harm done if someone puts in an invalid value.

If having a correct range becomes a critical concern in the future, we can add some kind of exception throwing to this implementation.

Yadunund: Thanks for the clarification!

PR #265, opened 2021-01-19
Address bug with task queue

Yadunund:
The current `TaskManager` implementation has a timer which periodically (every 1 second) calls `_begin_next_task()`. This function checks whether the `deployment time` for the front element of the task queue has passed and if so begins execution of the same (and then pops it from the queue).

However when a list of tasks with `start_time` of "now" is submitted, a potential race condition may arise:
Consider a fleet with two robots RobotA & RobotB.
A `BidProposal` for `TaskA` is received and is eventually queued in the `TaskManager` of `RobotA`. The task should technically begin execution but does not as the timer callback has not been called yet. (Still within 1s window)
A second `BidProposal` for `TaskB` is received. Here the task assignments will be replanned for a list comprising of the new task, `TaskB`, and any tasks still in the task queues for each robot in the fleet. So in this case the set of tasks `{TaskA, TaskB}` will be planned for. Assume the assignment returns such that `[{RobotB: {TaskB, TaskA}}, {RobotA:{}}]`, ie, RobotB is assigned to perform TaskB followed by TaskA and RobotA is not assigned any of the tasks. 
Now before the task queues of each robot are overwritten with the new assignments (ie, while waiting for the `DispatchResult` message from the Dispatcher node), the timer for RobotA executes (it has been 1s since "now") and so RobotA begins TaskA. (which has now technically been reassigned to RobotB). Then when the `DispatchResult` message arrives, the tasks queues are finally overwritten (too late). As a result `TaskA` will be performed twice: Once by `RobotA` right now, and again by `RobotB` after it finishes `TaskB`.

This PR introduces a stopgap fix for the problem described above.

### DIFF
--- a/rmf_fleet_adapter/src/rmf_fleet_adapter/TaskManager.cpp
+++ b/rmf_fleet_adapter/src/rmf_fleet_adapter/TaskManager.cpp
@@ -160,97 +160,103 @@ agv::ConstRobotContextPtr TaskManager::context() const
 void TaskManager::set_queue(
   const std::vector<TaskManager::Assignment>& assignments)
 {
-  std::lock_guard<std::mutex> guard(_mutex);
-  _queue.clear();
-
-  // We use dynamic cast to determine the type of request and then call the
-  // appropriate make(~) function to convert the request into a task
-  for (std::size_t i = 0; i < assignments.size(); ++i)
+  // We indent this block as _mutex is also locked in the _begin_next_task()
+  // function that is called at the end of this function.
   {
-    const auto& a = assignments[i];
-    auto start = _context->current_task_end_state().location();
-    if (i != 0)
-      start = assignments[i-1].state().location();
-    start.time(a.deployment_time());
-    rmf_task_msgs::msg::TaskType task_type_msg;
+    std::lock_guard<std::mutex> guard(_mutex);
+    _queue.clear();
 
-    if (const auto request =
-      std::dynamic_pointer_cast<const rmf_task::requests::Clean>(a.request()))
+    // We use dynamic cast to determine the type of request and then call the
+    // appropriate make(~) function to convert the request into a task
+    for (std::size_t i = 0; i < assignments.size(); ++i)
     {
-      task_type_msg.type = task_type_msg.TYPE_CLEAN;
-      auto task = rmf_fleet_adapter::tasks::make_clean(
-        request,
-        _context,
-        start,
-        a.deployment_time(),
-        a.state());
-      
-      _queue.push_back(task);
+      const auto& a = assignments[i];
+      auto start = _context->current_task_end_state().location();
+      if (i != 0)
+        start = assignments[i-1].state().location();
+      start.time(a.deployment_time());
+      rmf_task_msgs::msg::TaskType task_type_msg;
+
+      if (const auto request =
+        std::dynamic_pointer_cast<const rmf_task::requests::Clean>(a.request()))
+      {
+        task_type_msg.type = task_type_msg.TYPE_CLEAN;
+        auto task = rmf_fleet_adapter::tasks::make_clean(
+          request,
+          _context,
+          start,
+          a.deployment_time(),
+          a.state());
+        
+        _queue.push_back(task);
+      }
+
+      else if (const auto request =
+        std::dynamic_pointer_cast<const rmf_task::requests::ChargeBattery>(
+          a.request()))
+      {
+        task_type_msg.type = task_type_msg.TYPE_CHARGE_BATTERY;
+        const auto task = tasks::make_charge_battery(
+          request,
+          _context,
+          start,
+          a.deployment_time(),
+          a.state());
+
+        _queue.push_back(task);
+      }
+
+      else if (const auto request =
+        std::dynamic_pointer_cast<const rmf_task::requests::Delivery>(
+          a.request()))
+      {
+        task_type_msg.type = task_type_msg.TYPE_DELIVERY;
+        const auto task = tasks::make_delivery(
+          request,
+          _context,
+          start,
+          a.deployment_time(),
+          a.state());
+
+        _queue.push_back(task);
+      }
+
+      else if (const auto request =
+        std::dynamic_pointer_cast<const rmf_task::requests::Loop>(a.request()))
+      {
+        task_type_msg.type = task_type_msg.TYPE_LOOP;
+        const auto task = tasks::make_loop(
+          request,
+          _context,
+          start,
+          a.deployment_time(),
+          a.state());
+
+        _queue.push_back(task);
+      }
+
+      else
+      {
+        continue;
+      }
+
+      // publish queued task
+      rmf_task_msgs::msg::TaskSummary msg;
+      msg.task_id = _queue.back()->id();
+      msg.task_profile.task_id = _queue.back()->id();
+      msg.state = msg.STATE_QUEUED;
+      msg.robot_name = _context->name();
+      msg.fleet_name = _context->description().owner();
+      msg.task_profile.description.task_type = task_type_msg;
+      msg.start_time = rmf_traffic_ros2::convert(
+        _queue.back()->deployment_time());
+      msg.start_time = rmf_traffic_ros2::convert(
+        _queue.back()->finish_state().finish_time());
+      this->_context->node()->task_summary()->publish(msg);
     }
-
-    else if (const auto request =
-      std::dynamic_pointer_cast<const rmf_task::requests::ChargeBattery>(
-        a.request()))
-    {
-      task_type_msg.type = task_type_msg.TYPE_CHARGE_BATTERY;
-      const auto task = tasks::make_charge_battery(
-        request,
-        _context,
-        start,
-        a.deployment_time(),
-        a.state());
-
-      _queue.push_back(task);
-    }
-
-    else if (const auto request =
-      std::dynamic_pointer_cast<const rmf_task::requests::Delivery>(
-        a.request()))
-    {
-      task_type_msg.type = task_type_msg.TYPE_DELIVERY;
-      const auto task = tasks::make_delivery(
-        request,
-        _context,
-        start,
-        a.deployment_time(),
-        a.state());
-
-      _queue.push_back(task);
-    }
-
-    else if (const auto request =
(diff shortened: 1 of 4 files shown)


tanyouliang95 (approved):

mxgrey (commented):
Could you elaborate on how these changes address the problem that's stated in the PR description?

As far as I can tell, the material change is to call `_begin_next_task()` at the end of setting the queue, but I don't see how that changes anything for the described scenario.

Shouldn't the solution instead be to check each item in the task assignments and throw away any that have been started or that have already finished? Maybe I'm overlooking an important detail, but I don't see that happening in these changes.

Yadunund:
@mxgrey,

The only change as you pointed out, is to call `_begin_next_task()` at the end of the `set_queue()` function. I indented the rest of the `set_queue()` function to unlock the `_mutex` that is locked again by the `_begin_next_task()` function.

Regarding how this solves the problem:
It ensures that `RobotA` will immediately begin executing `TaskA` when it is added to the queue. (And not wait for the timer callback). Once this task is moved to "active" state, it will no longer be in the queue for `RobotA`. Thus, when a new `TaskB` request is received before the timer callback executes, the fleet adapter will solve the task allocation problem for the set of tasks `{TaskB}` as opposed to `{TaskB, TaskA}`. 

The method of checking each item in the assignments as proposed will not work as each robot has its own `TaskManager` with a task queue. So `RobotA` can be assigned `TaskA` and when the reshuffle happens, `RobotB` gets assigned `TaskA` and `TaskB`. But `RobotB`'s `TaskManager` has never encountered `TaskA` before (only `RobotA's` `TaskManager` has) so it will still treat it as a new assignment.

mxgrey:
> It ensures that RobotA will immediately begin executing TaskA when it is added to the queue.

This helps for this specific version of the race condition (which I suppose may be the most common manifestation of the race condition), but it doesn't help with other versions of the race condition, for example if `TaskA` won't be ready to start until later and gets reassigned because the adapter is awarded a bid that it had submitted before `TaskA` started to execute.

> The method of checking each item in the assignments as proposed will not work as each robot has its own TaskManager with a task queue.

Right, so that filtering needs to be done before attempting to assign the tasks. In fact, the *ideal* system design would be to check if any of the tasks mentioned in the bid have been started (or finished, although that's admittedly improbable), and do a replan if they have. If the replan is materially different than the original bid, then the dispatcher should be notified so it can reassign the task if necessary. But that's a much more complex system, so for now I'd suggest just doing a replan without notifying the dispatcher.

Yadunund:
Thanks for explaining that @mxgrey. I see how there are other situations where the race condition can arise. 

I'll make the changes to replan right before assigning tasks if there is a task that has started execution. 
However, wouldn't there still be a chance for a race condition during the time it takes to replan and re-assign? 🤔

mxgrey:
> However, wouldn't there still be a chance for a race condition during the time it takes to replan and re-assign?

If task starts and task assignments are processed sequentially (which I believe they should be since we're using a single-threaded executor) then there shouldn't be a problem. In the worst case scenario, maybe one of the tasks included in the replanning was started while the replanning was still being calculated. To deal with that case, when we get the replanned assignments, we should once again check that none of the tasks in the assignment have been started or finished (and trigger yet another replanning if one has, now with the started task removed).

Since the task allocation planning should be able to finish on a relatively small timescale (seconds at worst?) the probability of repeatedly needing to replan because of tasks being started is very very low. Additionally, each time that edge case occurs, there are fewer tasks that need to be replanned, so the replanning would get faster and faster until (in the most extreme case) only one task remains.

What I'm less clear on is how the behavior might play out if a new bid award comes in during the replanning. Does the dispatcher guarantee that each bid has to finish before the next bid can begin. E.g. if the dispatcher receives `TaskRequestA` and `TaskRequestB` (almost) simultaneously, is it guaranteed to finish assigning `TaskRequestA` before it begins the bid for `TaskRequestB`, or will it open bids for `TaskRequestA` and `TaskRequestB` in parallel? If it's the latter, then I'll need to look much more carefully at how the bid awards are being handled in general.

Yadunund:
Gotcha. Will implement some form of recursive checking of task assignments.

> Does the dispatcher guarantee that each bid has to finish before the next bid can begin.
Correct me if i'm wrong @tanyouliang95 , but I believe the bid for `TaskRequestB` will only be sent **after** the `TaskRequestA` has be awarded.

tanyouliang95:
@Yadunund  Yes, the dispatcher does issue a `DispatchRequest` before publishing a new `BidNotice`. But since they are different topics, I'm not quite sure the msgs sequence is always guaranteed.

mxgrey:
> But since they are different topics, I'm not quite sure the msgs sequence is always guaranteed.

Taking this into account, the most robust design possible would be to wait for an acknowledgment from the award recipient before beginning the next task bid.

Yadunund:
Sure thing. We can have the fleet adapter publish a `DispatchAcknowledgement` message after it updates its assignments for an awarded bid. This message can contain a `TaskID` and a `boolean` field indicating whether the operation was successful.

Alternatively, the dispatcher can wait for the `TaskSummary` message with state `Queued` for the awarded task before beginning the bid for the next task.

Yadunund:
@mxgrey , 
I've updated the bug fix to check for validity of assignments before overwriting the queues in the `TaskManagers`. If the assignment set is no longer valid (one of the tasks has started execution), the assignments will be re-planned and re-validated. Queues are overwritten only when re-validation passes. If not, the process of re-planning and re-validating will be repeated until the assignments are valid.

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
```diff
@@ -557,63 +492,124 @@ void FleetUpdateHandle::Implementation::bid_notice_cb(
...
+        if (!replan_results)
+        {
+          RCLCPP_WARN(
+            node->get_logger(),
+            "Unable to replan assignments for task_id:[%s]", id.c_str());
+          return;
+        }
+        assignments = replan_results.value();
+        // We re-check the assignments as a task could have started while replanning
+        valid_assignments = is_valid_assignments(assignments);
```
mxgrey, line 561: This whole function is being called by the ROS2 executor, which means it's running on the main rxcpp worker. There isn't any risk that the assignments would be invalidated during the time it takes to replan, because no new tasks will be started while the planning is happening.

On that note, it might not be good that the replanning is blocking the main thread. If tests have been going okay, then we can leave this as-is for now, but a better non-blocking solution would be to run the replanning on a separate worker and then deliver the result back to the main worker. Let's at least leave a note here to improve this so that it doesn't block.

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
```diff
@@ -557,63 +492,124 @@ void FleetUpdateHandle::Implementation::bid_notice_cb(
...
+  }
 
+  else if (msg->method == DispatchRequest::CANCEL)
+  {
+    // TODO(YV)
+  }
+
+  else
+  {
     return;
```
mxgrey, line 587: We should probably return a warning message saying that we received an unsupported dispatch request type.

I would also suggest that we put a "Not implemented yet" warning message for the `CANCEL` case above.

mxgrey (changes requested):
I have a few small change requests, some of which are just to add notes to improve the implementation in the future when time allows.

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
Yadunund: Done https://github.com/osrf/rmf_core/pull/265/commits/347e4bff842986b6ba7dc0d653bc7743f53b57b2

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
Yadunund: Thank you for explaining this. I've removed the "re-check" for the assignments and added notes explaining the above. https://github.com/osrf/rmf_core/pull/265/commits/347e4bff842986b6ba7dc0d653bc7743f53b57b2

So far in my testing, I have not encountered any delays due to the blocking nature of this function. Will continue to keep an eye out for this and address the issue in the future.

rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp
```diff
@@ -425,90 +426,14 @@ void FleetUpdateHandle::Implementation::bid_notice_cb(
...
-    else
-    {
-      RCLCPP_ERROR(
-        node->get_logger(),
-        "[TaskPlanner] Failed to compute assignments for task_id:[%s]",
-        id.c_str());
-    }
+  const auto allocation_result = allocate_tasks(new_request);
 
+  if (!allocation_result)
```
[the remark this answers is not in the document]
Yadunund: Thanks for the heads up. https://github.com/osrf/rmf_core/pull/265/commits/347e4bff842986b6ba7dc0d653bc7743f53b57b2

rmf_fleet_adapter/src/rmf_fleet_adapter/TaskManager.cpp
```diff
@@ -479,4 +493,20 @@ void TaskManager::retreat_to_charger()
...
+//==============================================================================
+const std::vector<std::string>& TaskManager::get_executed_tasks() const
+{
+  return _executed_task_registry;
+}
+
+//==============================================================================
+void TaskManager::_register_executed_task(const std::string& id)
+{
+  if (_executed_task_registry.size() == 100)
```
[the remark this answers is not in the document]
Yadunund: Agreed that the condition is arbitrary. I like the idea of saving a time stamp for when the job is completed and then cull the entries accordingly. Have left a note for the next iteration. https://github.com/osrf/rmf_core/pull/265/commits/347e4bff842986b6ba7dc0d653bc7743f53b57b2

mxgrey (approved):

PR #278, opened 2021-02-02
Priority in Task Allocation

Yadunund:
This PR aims to introduce the notion of **priority** in the task allocation framework.

In the proposed method, abstract class `rmf_task::Request` has a new `priority` attribute which takes a boolean value (`true` if high priority). Then, in `rmf_task::agv::TaskPlanner`, the total cost of a node is multiplied by a `priority_penalty` (arbitrary large number) if the order of assignments is invalid. Two checks are performed inside `valid_assignment_priority()`: 1) Check if any agent is not assigned a high priority task while other agents are assigned more than one 2) Within the assignments for an agent, check if the sequence of assigned tasks contains a high priority task after a low priority one. The entry for this function is [here](https://github.com/osrf/rmf_core/blob/09cfbcd67e92127f30d0b5afce5e6753c2eed26e/rmf_task/src/rmf_task/agv/TaskPlanner.cpp#L1030-L1034). The proposed method works and some tests are included to prove the same.

Opening this PR as a draft to discuss this approach or receive advice on other approaches. 

Once the suitable approach is confirmed, I can make changes to `rmf_task_msgs` to include a priority field and relevant changes to `FleetUpdateHandle.cpp`


rmf_task/include/rmf_task/Request.hpp
```diff
@@ -39,6 +39,9 @@ class Request
   /// Get the id of the task
   virtual std::string id() const = 0;
 
+  /// Get the priority of the task. High priority tasks should return true.
+  virtual bool priority() const = 0;
```
mxgrey, line 43: I might recommend refactoring `Request` to be a concrete class instead of a pure abstract interface.

We could have `id`, `earliest_start_time`, and `priority` be concrete fields. `id` and `earliest_start_time` would keep their current value types (`std::string` and `rmf_traffic::Time`) while `priority` could be some kind of `std::unique_ptr<Priority>` or `std::shared_ptr<Priority>` field where `Priority` is a pure abstract interface that gets generated based on whatever the "prioritization scheme" is for the cost function.

Meanwhile we could refactor the cost calculations for the planner into its own abstract interface. Different prioritization schemes can be chosen by selecting the cost calculation implementation. That cost calculation implementation would be responsible for providing the `Priority` objects that get held by the `Request` objects. For cost calculators that don't have any prioritization schemes, the `priority` field can just be left as a `nullptr`.

Yadunund: I've managed to refactor `Request` as suggested but am having trouble with the cost calculation abstraction and implementation. The calculation of [assignment cost](https://github.com/osrf/rmf_core/blob/b50b5cd16eac925fa107b7c4b9748a31aa696005/rmf_task/src/rmf_task/agv/TaskPlanner.cpp#L867-L878) and [heuristic cost](https://github.com/osrf/rmf_core/blob/b50b5cd16eac925fa107b7c4b9748a31aa696005/rmf_task/src/rmf_task/agv/TaskPlanner.cpp#L880-L931) (and therefore total cost) depend heavily on the internal [Node definition](https://github.com/osrf/rmf_core/blob/b50b5cd16eac925fa107b7c4b9748a31aa696005/rmf_task/src/rmf_task/agv/TaskPlanner.cpp#L431-L495). Is there a way to proceed without making the `Node` definition part of the public API?

I considered templating the `CostCalculator` class as below. But then the sample implementation of this class (`SimpleCostCalculator`) would still need to define `Node` so that downstream packages can instantiate an object of `SimpleCostCalculator`.

```
template<typename Node>
class CostCalculator
{
public:

  /// Compute the total cost while factoring in the prioritization scheme for a node
  virtual double compute_cost(const Node& n) const = 0;

  /// Generate a priority object based on the prioritization scheme of the CostCalculator
  virtual ConstPriorityPtr priority(std::size_t value) const = 0;

  virtual ~CostCalculator() = default;
};
```
Please let me know if i'm misunderstanding something.

mxgrey: That's a great point which I hadn't completely thought through.

I'd say for now let's keep the cost calculation abstraction internal so we don't have to commit the `Node` definitions to the public API. That might look something like this:

```
// Forward declaration of these abstract interfaces in a public header.
// The definitions of these abstract interfaces will remain internal.
class CostCalculator;
class Priority;

/// This is for a binary prioritization scheme. Tasks are assigned either high priority or low priority.
class HighLowPriority
{
public:

  /// Use these to assign the task priority
  static std::shared_ptr<Priority> make_low_priority();
  static std::shared_ptr<Priority> make_high_priority();

  /// Use this to give the appropriate cost calculator to the planner
  static std::shared_ptr<CostCalculator> make_cost_calculator();
};
```

The classes that implement the abstract interfaces `CostCalculator` and `Priority` can remain internal implementation details.

The one weakness I see with the approach that I'm suggesting is it can't stop users from mixing together `Priority` implementations with incompatible `CostCalculator` implementations. It may be worth thinking about how to make the API more narrow so that users can't do the wrong thing. I think the approach of forward declaring the abstract interfaces will still be useful for that, though.

Yadunund: Thanks for the advice- this is a really neat approach.

mxgrey: Maybe `class BinaryPriority` would be a better name than `class HighLowPriority`.

Yadunund: I have refactored the package as suggested.
I renamed the above class to `BinaryPriorityScheme`. `Priority` and `CostCalculator` abstract classes are now defined as internal implementation details along with their respective sample implementations `BinaryPriority` and `BinaryPriorityCostCalculator`.
If things look okay I can proceed with updating the fleet adapter side of things.

I plan to introduce a new `Priority.msg` in `rmf_task_msgs` with a single field `uint64 value`. Then the [TaskProfile](https://github.com/osrf/rmf_core/blob/master/rmf_task_msgs/msg/TaskProfile.msg) will be updated with a new `Priority priority` field. I'm hoping having a single `value` field will be able to satisfy multiple priority schemes. Please let me know if this makes sense.

rmf_task/include/rmf_task/BinaryPriorityScheme.hpp
```diff
@@ -0,0 +1,46 @@
...
+*/
+
+#ifndef RMF_TASK__BINARYPRIORITYSCHEME_HPP
+#define RMF_TASK__BINARYPRIORITYSCHEME_HPP
+
+#include <memory>
+
+namespace rmf_task {
+
+// Forward declare abstract interfaces. The definition will remain internal
```
mxgrey, line 25: I recommend putting these forward declarations and type aliases into their own headers. Probably a `Priority.hpp` and `CostCalculator.hpp`.

I realize it probably seems unnecessary for some plain forward declarations, but it's best to have a single source of truth for the names of these classes.

rmf_task/include/rmf_task/Request.hpp
```diff
@@ -34,25 +35,45 @@ class Request
...
+  ///
+  /// \param[in] earliest_start_time
+  ///   The earliest time this task should begin execution. This is usually the
+  ///   requested start time for the task.
+  ///
+  /// \param[in] priority
+  ///   The priority for this task. This is provided by the CostCalculator. For
+  ///   requests that do not have any priority this is a nullptr
+  Request(
+    std::string& id,
```
mxgrey, line 51:
```suggestion
    const std::string& id,
```

rmf_task/include/rmf_task/Request.hpp
```diff
@@ -34,25 +35,45 @@ class Request
...
+  /// Get the earliest start time that this task may begin
+  rmf_traffic::Time earliest_start_time() const;
+
+  /// Get the priority scheme of this task
+  ConstPriorityPtr priority() const;
 
   /// Estimate the state of the robot when the task is finished along with the
   /// time the robot has to wait before commencing the task
-  virtual rmf_utils::optional<Estimate> estimate_finish(
+  virtual std::optional<Estimate> estimate_finish(
```
mxgrey, line 66: We should never ever ever mix together concrete types and abstract interfaces.

Let's define a new class to be the abstract interface for these functions, and then add it as a field to the `Request` class. For example:

```
class Request
{
public:
  class Description
  {
  public:
    virtual std::optional<Estimate> estimate_finish(...) const = 0;
    virtual rmf_traffic::Duration invariant_duration() const = 0;
    virtual ~Type() = default;
  };
  using DescriptionPtr = std::shared_ptr<const Description>;

  Request(
    const std::string& id,
    rmf_traffic::Time earliest_start_time,
    ConstPriorityPtr priority,
    DescriptionPtr description);
};
```

rmf_task/include/rmf_task/agv/TaskPlanner.hpp
```diff
@@ -95,7 +94,14 @@ class TaskPlanner
     std::shared_ptr<rmf_traffic::agv::Planner> planner() const;
 
     /// Set the planner
-    Configuration& planner(std::shared_ptr<rmf_traffic::agv::Planner>);
+    Configuration& planner(std::shared_ptr<rmf_traffic::agv::Planner> planner);
+
+    /// Get the CostCalculator
+    std::shared_ptr<rmf_task::CostCalculator> cost_calculator() const;
```
mxgrey, line 100:
```suggestion
    const std::shared_ptr<rmf_task::CostCalculator>& cost_calculator() const;
```

We can also consider changing the other accessor functions to return their `std::shared_ptr` by const-reference.

rmf_task/include/rmf_task/requests/ChargeBattery.hpp
```diff
@@ -67,7 +64,11 @@ class ChargeBattery : public rmf_task::Request
 
   class Implementation;
 private:
-  ChargeBattery();
+  ChargeBattery(
```
mxgrey, line 67: We'll want to change this class (and the other request classes) to implement `rmf_task::Request::Description` instead of `Request`.

rmf_task/src/rmf_task/CostCalculator.cpp
```diff
@@ -0,0 +1,325 @@
...
+  struct element { double start; double end; };
+  std::vector<std::vector<element>> _stacks;
+};
+
+} // anonymous namespace
+
+namespace rmf_task {
+
+//==============================================================================
+class BinaryPriorityCostCalculator::Implementation
```
mxgrey, line 92: I recommend moving the `BinaryPriorityCostCalculator` into its own `BinaryPriorityCostCalculator.cpp`. As we potentially add more cost calculation schemes, we'll get better compile times and cleaner code organization if they're in separate translation units.

mxgrey: The corollary to this is that the definition of `InvariantHeuristicQueue` will have to go into an internal header. And then the definitions of its member functions would ideally go into `InvariantHeurisitcQueue.cpp`. I apologize for the amount of copy/pasting this will require.

rmf_task/src/rmf_task/Priority.hpp
```diff
@@ -0,0 +1,59 @@
...
+
+  virtual ~Priority() = default;
+};
+
+using PriorityPtr = std::shared_ptr<Priority>;
+using ConstPriorityPtr = std::shared_ptr<const Priority>;
+
+//==============================================================================
+// Sample implementation for binary prioritization scheme
+class BinaryPriority : public Priority
```
mxgrey, line 43: Let's define `BinaryPriority` in its own header file for the sake of cleaner code organization.

rmf_task/src/rmf_task/agv/TaskPlanner.cpp
```diff
@@ -123,6 +120,22 @@ auto TaskPlanner::Configuration::planner(
...
+{
+  return _pimpl->cost_calculator;
+}
+
+//==============================================================================
+auto TaskPlanner::Configuration::cost_calculator(
+  std::shared_ptr<rmf_task::CostCalculator> cost_calculator) -> Configuration&
+{
+  if (cost_calculator)
+    _pimpl->cost_calculator = cost_calculator;
```
mxgrey, line 135: I might suggest allowing this to have a `nullptr` value and then have the planner automatically select a default if the user provided a `nulltpr`.

In cases where `nullptr` inputs are unacceptable, I think it's better to throw an exception rather than silently do nothing.

mxgrey (changes requested):
I think this is shaping up really nicely!

I do have some inline feedback on how to smooth out some rough edges in the current approach to the API.

rmf_task/include/rmf_task/Request.hpp
Yadunund: I'm having trouble refactoring as suggested as the `estimate_finish()` function internally requires the `estimate_start_time` field. Would it be okay to move `earliest_start_time()` and `id()` into `Description` as virtual functions? I would like to move `id()` as well since this field is sometimes populated internally as in the case of automatically added tasks such as `ChargeBattery`. 
Proposed design:
```
class Request
{
public:
  class Description
  {
  public:

    /// The unique id for this request
    virtual const std::string& id() const = 0;

    /// Get the earliest time that this request may begin
    rmf_traffic::Time earliest_start_time() const = 0;

    /// Estimate the state of the robot when the request is finished along with the
    /// time the robot has to wait before commencing the request
    virtual std::optional<Estimate> estimate_finish(
      const agv::State& initial_state,
      const agv::Constraints& task_planning_constraints,
      const std::shared_ptr<EstimateCache> estimate_cache) const = 0;
    
    /// Estimate the invariant component of the request's duration
    virtual rmf_traffic::Duration invariant_duration() const = 0;

    virtual ~Description() = default;
  };
  using DescriptionPtr = std::shared_ptr<Description>;

  /// Constructor
  ///
  /// \param[in] earliest_start_time
  ///   The earliest time this request should begin execution. This is usually the
  ///   requested start time for the request.
  ///
  /// \param[in] priority
  ///   The priority for this request. This is provided by the Priority Scheme. For
  ///   requests that do not have any priority this is a nullptr.
  ///
  /// \param[in] description
  ///   The description for this request
  Request(
    ConstPriorityPtr priority,
    DescriptionPtr description);

  /// Get the priority of this request
  ConstPriorityPtr priority() const;

  /// Get the description of this request
  const DescriptionPtr& description() const;

  class Implementation;
private:
  rmf_utils::impl_ptr<Implementation> _pimpl;
};
```

mxgrey: I think the abstraction that I proposed is the right way to do it.

I can think of two ways to deal with the earliest start time:

1. Design the factory functions for the various `Request::Description` implementations to take in an `earliest_start_time` argument, and have the `Request::Description` internally save its own copy for that field. We'd be passing the `earliest_start_time` as an argument to both the constructor of `Request` and the factory functions of `Request::Description`, so there's some duplication of truth, but I think that's acceptable in this case.
2. Have the `estimate_finish` function accept an `earliest_start_time` argument. This may be preferable if the `Request::Description` objects don't need to assume that the `earliest_start_time` will always be the same value, but it may be problematic if we ever want to allow them to pre-compute and/or cache anything based on the earliest start time of the task. They should still be able to do caching by maintaining a `std::unordered_map<rmf_traffic::Time, ...>` cache, it'll just be less convenient.

> I would like to move id() as well since this field is sometimes populated internally as in the case of automatically added tasks such as ChargeBattery.

I don't see any reason the ID can't be automatically generated for the `Request` internally when automatically generating a battery charging task. You would just have a battery-specific factory function that returns a `Request` object, and that factory function can generate the ID and use it to construct the `Request` object before returning that `Request` to you.

Overall I don't think there's a compelling reason to have those fields be virtual, since I can't conceive of a situation where their behavior would be implementation dependent. The start time and ID fields should both be values that get set during `Request` construction, and their accessor functions should just return whatever that original value was.

Yadunund: Thanks for the detailed feedback. I will go with the first suggestion and store a copy of the `earliest_start_time` inside `Request::Description`. I didn't consider having two factory functions, one for generating a request-specific `Request` objects and the other for the request-specific `Request::Description` objects. I see how its possible to keep these fields concrete.

rmf_task/include/rmf_task/BinaryPriorityScheme.hpp
Yadunund: Done https://github.com/osrf/rmf_core/pull/278/commits/cced88c75d76dbcef2d39268700d388cca86146c

rmf_task/include/rmf_task/Request.hpp
Yadunund: I've updated the APIs accordingly
https://github.com/osrf/rmf_core/pull/278/commits/55d01babe67efd82f48ebf5dd1f684c6ca187b88
https://github.com/osrf/rmf_core/pull/278/commits/ed42b6f35c46dbce3d52ede0748bb762b076a4c8
https://github.com/osrf/rmf_core/pull/278/commits/e0a7c852e0a822b7bad4b913e236b7b2733e339b

rmf_task/include/rmf_task/agv/TaskPlanner.hpp
Yadunund: Done https://github.com/osrf/rmf_core/pull/278/commits/814174cffd80a473f649c39a5d3024e9fc7ce8bb

rmf_task/src/rmf_task/CostCalculator.cpp
Yadunund: Done https://github.com/osrf/rmf_core/pull/278/commits/b7d22e17ef093fa6fd62d77b62ab5353f24c1630

rmf_task/src/rmf_task/Priority.hpp
Yadunund: Done https://github.com/osrf/rmf_core/pull/278/commits/b7d22e17ef093fa6fd62d77b62ab5353f24c1630

rmf_task/src/rmf_task/agv/TaskPlanner.cpp
Yadunund: Done https://github.com/osrf/rmf_core/pull/278/commits/18beb68180600b855b48acdb86176e2ed85d6855

rmf_task/src/rmf_task/BinaryPriorityScheme.cpp
```diff
@@ -0,0 +1,43 @@
...
+
+#include "BinaryPriority.hpp"
+#include "BinaryPriorityCostCalculator.hpp"
+
+namespace rmf_task {
+
+//==============================================================================
+std::shared_ptr<Priority> BinaryPriorityScheme::make_low_priority()
+{
+  return nullptr;
```
mxgrey, line 28: I feel a bit conflicted about this.

On one hand, it makes sense to use `nullptr` since it would be more efficient than allocating an object to the heap and dynamically casting it to view the value.

On the other hand, it could be confusing for users to receive a `nullptr` when they've asked for an object to be created. It might give the impression that something failed inside the `rmf_task` library.

Let's add documentation to the `make_low_priority` function to explain that the current implementation of this function simply returns a `nullptr`, and that's the normal, expected behavior until further notice.

rmf_task/src/rmf_task/Priority.hpp
```diff
@@ -0,0 +1,43 @@
...
+
+namespace rmf_task {
+
+/// Implement this for a new prioritization scheme.
+class Priority
+{
+public:
+
+  /// Get the value of this priority
+  virtual std::size_t value() const = 0;
```
mxgrey, line 33: What I'm about to say might sound strange, but we don't need to make any virtual functions for this "interface" class.

The `Priority` class can just be a blank base class, and the `CostCalculator` implementations can just dynamic cast the `Priority` objects into whatever type the `CostCalculator` is compatible with.

rmf_task/src/rmf_task/Priority.hpp
```diff
@@ -0,0 +1,43 @@
...
+
+  virtual ~Priority() = default;
```
mxgrey, line 35: After removing the virtual function above, we can change this to

```
virtual ~Priority() = 0
```

which is a C++ syntax trick to make this class a pure abstract class without having any pure virtual functions. Being a pure abstract class means it's impossible to create a `Priority` object directly; you can only create an object of a class derived from `Priority`.

rmf_task/src/rmf_task/Request.cpp
```diff
@@ -0,0 +1,73 @@
...
+Request::Request(
+  const std::string& id,
+  rmf_traffic::Time earliest_start_time,
+  ConstPriorityPtr priority,
+  DescriptionPtr description)
+: _pimpl(rmf_utils::make_impl<Implementation>(
+      Implementation {
+          id,
+          earliest_start_time,
+          priority,
```
mxgrey, line 42: Nitpick: Putting `std::move(~)` on `priority` and `description` will save some reference counting overhead.

rmf_task/src/rmf_task/agv/TaskPlanner.cpp
```diff
@@ -746,6 +344,8 @@ class TaskPlanner::Implementation
 
   std::shared_ptr<Configuration> config;
   std::shared_ptr<EstimateCache> estimate_cache;
+  bool check_priority = false;
```
mxgrey, line 347: The way this `check_priority` flag is being used, it seems rather specific to the `BinaryPriorityScheme` implementation.

Since this is an internal implementation detail, it's fine to leave it as-is for now, but we'll probably want to rework this design if/when we start to support alternative prioritization or cost schemes.

mxgrey (changes requested):
This is looking really good! I just have a few minor nitpicks remaining, mostly related to interface/implementation semantics.

rmf_task/src/rmf_task/BinaryPriority.hpp
```diff
@@ -0,0 +1,42 @@
...
+public:
+
+  /// Constructor
+  BinaryPriority(std::size_t value);
+
+  /// Documentation inherited
+  std::size_t value() const final;
+  class Implementation;
+private:
+  rmf_utils::impl_ptr<Implementation> _pimpl;
```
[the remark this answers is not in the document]
Yadunund: Done https://github.com/osrf/rmf_core/pull/278/commits/68ec61f4cdee5eae420a912d29b750b00acb1826

rmf_task/src/rmf_task/BinaryPriorityScheme.cpp
Yadunund: Have added documentation https://github.com/osrf/rmf_core/pull/278/commits/68ec61f4cdee5eae420a912d29b750b00acb1826

rmf_task/src/rmf_task/Priority.hpp
Yadunund: Thanks for explaining this. The compiler required me to add a function body for the pure virtual destructor so I've added that in `Priority.cpp` (that is the only contents of this file). 

https://github.com/osrf/rmf_core/pull/278/commits/68ec61f4cdee5eae420a912d29b750b00acb1826

rmf_task/src/rmf_task/Request.cpp
Yadunund: Done https://github.com/osrf/rmf_core/pull/278/commits/68ec61f4cdee5eae420a912d29b750b00acb1826

Yadunund:
Thanks for all the review suggestions @mxgrey. I think I've addressed the remaining points.

mxgrey (approved):
Looks great!

PR #298, opened 2021-02-16
Alternative StubbornNegotiator that shares ownership of participant

mxgrey:
The requirement that the `StubbornNegotiator` will always have a very short lifecycle may be too confusing and restrictive for some users, so this adds an alternative constructor that accepts a shared reference to the participant. This will help in situations where the `StubbornNegotiator` has a long lifecycle and needs to be tied to a participant which may move around between different owners.


mxgrey:
This is a very minor and self-contained feature addition, so I'll be merging this without review.

PR #299, opened 2021-02-17
Fix updating bugs in traffic light implementation

mxgrey:
While preparing for the xnergy demo, some bugs were identified in the traffic light implementation, which caused the schedule to not get updated correctly in some edge cases.

This PR fixes the issues that were identified from the demo work.


codebot (approved):
tested and verified on real hardware :+1:

